PR #1, opened 2015-10-05
Responded to questoins in challenge.js

TeoPaolucci:
No external sources used.

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -19,7 +19,18 @@
   // - park
 
 // Answer Starts Here
-
+// 'Calendar'
+// Description : Our app manages events on a calendar.
+// Every 'calendar' has
+// - a Year
+// - a Month (each Year has these)
+// - a Week (each Month has these)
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -19,7 +19,18 @@
...
 
 // Answer Starts Here
-
+// 'Calendar'
+// Description : Our app manages events on a calendar.
+// Every 'calendar' has
+// - a Year
+// - a Month (each Year has these)
+// - a Week (each Month has these)
+// - a Day (each Week has these)
```
GA-MEB, line 28: Calendar is a reasonable abstraction - can you name two others?

Or are Year, Month, Week, and Day your other abstractions?

GA-MEB:
3.5 of 4

I took off 0.5 from Q1, but I could be persuaded to add it back if you can demonstrate two other good abstractions for that problem. 

Overall, very nice work!

PR #3, opened 2015-10-05
Added my responses into challenge.js

gotwilk890:


lib/challenge.js
```diff
@@ -20,6 +20,16 @@
...
 
+'Calandar'
+Description: our web browser is a calandar app
+Every Calandar has
+- month
+- day
+- events
+Every 'Calandar' can
+- schedule events
+- print out itself to hang up.
```
GA-MEB, line 31: This is a good abstraction. Can you think of two other reasonable abstractions?

GA-MEB:
3.5 of 4 (-0.5 on Q1)

Nice work. Let me know if you have questions about Q1.

PR #4, opened 2015-10-05
Finish Assessment

dhuddell:
Add answers
Finish assessment

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -19,7 +19,20 @@
   // - park
 
 // Answer Starts Here
+Every 'event' has
+- a date
+- a description
+- a location
+Every 'event' can
+- remind
+- request other users
 
+Every 'user' has
+- a name
+- an email
+Every 'user' can
+- schedule events
+- RSVP to events
 // Answer Ends Here
 
 // 2. Create an Object literal that lines up with the following description. Store it in the variable 'pet_owner', below. Be sure to give it reasonable values for each of its properties.
@@ -33,7 +46,10 @@
 var pet_owner;
 
 // Answer Starts Here
-
+pet_owner = {
+  name : "Jimmy Jo Johston",
+  address : "1 Apple St."
+}
 // Answer Ends Here
 
 // 3. Create an Object literal that lines up with the following description. Store it in the variable `some_pet`, below.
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -19,7 +19,20 @@
...
+Every 'event' can
+- remind
+- request other users
+
+Every 'user' has
+- a name
+- an email
+Every 'user' can
+- schedule events
+- RSVP to events
```
GA-MEB, line 35: These look great! Can you think of one more?

lib/challenge.js
```diff
@@ -51,7 +67,12 @@ var pet_owner;
 var some_pet;
 
 // Answer Starts Here
-
+some_pet = {
+  name : "Miley Cyrus",
+  species : "Capybara",
+  breed : "Scottish Biter",
+  noise : "Harumph"
+}
```
GA-MEB, line 75: Whoops - missing the `makeNoise` method. `somePet` should look something like this:

``` javascript
somePet = {
  name : "Miley Cyrus",
  species : "Capybara",
  breed : "Scottish Biter",
  noise : "Harumph",
  makeNoise : function(){
    return this.noise;
  }
};
```

dhuddell, line 75: I guess I made a mistake in taking it out, bc we have the constructor function after it (odd order) which I did include it in.

lib/challenge.js
```diff
@@ -60,7 +81,15 @@ var some_pet;
 
 var Pet;
 // Answer Starts Here
-
+Pet = function(name, species, breed, noise){
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+  this.makeNoise = function(){
```
GA-MEB, line 89: In this case, since our Constructor function will be making many identical objects, we don't want it to make a brand new copy of our `makeNoise` method each time. To avoid this, we can use the prototype of Pet, and define all of our methods there; since all individual pets will have access to Pet.prototype, they can still each use all those methods, but it will use a lot less memory.

``` javascript
  this.name = name;
  this.species = species;
  this.breed = breed;
  this.noise = noise;
}

  return this.noise;
};
```

GA-MEB:
2.5 of 4 (-0.5 on Q1, Q3, Q3)

Good work. I know that 2.5/4 looks not so great, but you were pretty close on all three of those questions. Feel free to resubmit if you want, but don't feel like you have to either.

Let me know if you have any questions on the prototype stuff!

PR #6, opened 2015-10-05
"Answered challenge questions"

jcmeyer10:
Utilized Jeff's repl.it to confirm prototype syntax and place in Object Literal creation function.

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -97,9 +97,7 @@ var Pet = function getPetInfo(name, species, breed, noise) = {
 var Pet = New getPetInfo();
 
 Pet.prototype.makeNoise = function makeNoise() {
-  if (species === "Dog") {
-      return "Bark"
-    }
+      return this.noise
   };
 };
 
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -20,6 +20,18 @@
 
 // Answer Starts Here
 
+'Calendar'
+Decription: Our app keeps track of events for day.
+1) Every Calendar has:
+a) Dates
+b) Hour markers
+c) Months
+
+2) Every calendar can:
+a) Add events
+b) Alert before an event
+c) Repeat events at specified intervals
+
 // Answer Ends Here
 
 // 2. Create an Object literal that lines up with the following description. Store it in the variable 'pet_owner', below. Be sure to give it reasonable values for each of its properties.
@@ -34,6 +46,10 @@ var pet_owner;
 
 // Answer Starts Here
 
+var pet_owner = {
+  name: "Jeff Meyer"
+  email: "160 Quincy Shore Drive, Quincy MA 02171"
+};
 // Answer Ends Here
 
 // 3. Create an Object literal that lines up with the following description. Store it in the variable `some_pet`, below.
@@ -51,7 +67,17 @@ var pet_owner;
 var some_pet;
 
 // Answer Starts Here
-
+var some_pet = {
+  name: "Rupert"
+  species: "Dog"
+  breed: "Goldendoodle"
+  noise: "Bark"
+  makeNoise: function makeNoise = {
+    if (species === "Dog") {
+      return "Bark"
+    }
+  };
+};
 // Answer Ends Here
 
 // 4. Create a constructor function for Pet objects, as described above. The constructor function should be responseible for setting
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -20,6 +20,18 @@
...
+1) Every Calendar has:
+a) Dates
+b) Hour markers
+c) Months
+
+2) Every calendar can:
+a) Add events
+b) Alert before an event
+c) Repeat events at specified intervals
+
```
GA-MEB, line 34: This is a pretty reasonable abstraction. Can you think of two more?

lib/challenge.js
```diff
@@ -51,7 +67,17 @@ var pet_owner;
...
-
+var some_pet = {
+  name: "Rupert"
+  species: "Dog"
+  breed: "Goldendoodle"
+  noise: "Bark"
+  makeNoise: function makeNoise = {
+    if (species === "Dog") {
+      return "Bark"
+    }
```
GA-MEB, line 78: Mmm... not quite what we were looking for. What if `some_pet.noise` isn't "bark" but, rather, "woof"? `makeNoise` should reference the `noise` property of `some_pet`. In other words,

``` javascript
makeNoise : function() {
  return this.noise;
}
```

lib/challenge.js
```diff
@@ -51,7 +67,17 @@ var pet_owner;
 var some_pet;
 
 // Answer Starts Here
-
+var some_pet = {
+  name: "Rupert"
+  species: "Dog"
+  breed: "Goldendoodle"
+  noise: "Bark"
+  makeNoise: function makeNoise = {
```
GA-MEB, line 75: Don't forget the `()`!

lib/challenge.js
```diff
@@ -60,6 +86,20 @@ var some_pet;
 
 var Pet;
 // Answer Starts Here
+var Pet = function getPetInfo(name, species, breed, noise) = {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise
+  this.makeNoise = makeNoise();
```
GA-MEB, line 94: This line doesn't mean what you think it means. It's creating a `makeNoise` attribute for all pets, and setting the value to the result of some function that exists elsewhere called `makeNoise`.

lib/challenge.js
```diff
@@ -60,6 +86,20 @@ var some_pet;
...
+}
+
+var Pet = New getPetInfo();
```
GA-MEB, line 97: `new` is a keyword, and should not be capitalized. Also, although they're technically the same function in your code, the convention would be to use `Pet` rather than `getPetInfo`. 

Additionally, you shouldn't be assigning the result back into `Pet`. First, and most importantly, it overwrites your constructor function. And secondly, it goes against convention to capitalize anything but a constructor function; the result of `new` should not be another constructor function, but rather a new object.

lib/challenge.js
```diff
@@ -60,6 +86,20 @@ var some_pet;
...
+
+Pet.prototype.makeNoise = function makeNoise() {
```
GA-MEB, line 99: This part is correct!

jcmeyer10, line 99: Ah, ok, if you have a few minutes later, I would like to go over it just to
make sure I get it right.

On Mon, Oct 5, 2015 at 3:39 PM, Matt Brendzel notifications@github.com
wrote:

> In lib/challenge.js
> https://github.com/ga-wdi-boston/js-objects-intro-assessment/pull/6#discussion_r41187305
> :
> 
> > @@ -60,6 +86,20 @@ var some_pet;
> > 
> >  var Pet;
> >  // Answer Starts Here
> > +var Pet = function getPetInfo(name, species, breed, noise) = {
> > -  this.name = name;
> > -  this.species = species;
> > -  this.breed = breed;
> > -  this.noise = noise
> > -  this.makeNoise = makeNoise();
> >   +}
> >   +
> >   +var Pet = New getPetInfo();
> >   +
> >   +Pet.prototype.makeNoise = function makeNoise() {
> 
> This part is correct!
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/ga-wdi-boston/js-objects-intro-assessment/pull/6/files#r41187305
> .

## 

Regards,
Jeff Meyer

857-206-3572

lib/challenge.js
GA-MEB, line 97: Here's some correct syntax:
`var somePet = new Pet();`

lib/challenge.js
GA-MEB, line 99: Sure thing.

GA-MEB:
1.5 of 4

Let's work through this together later today!

PR #7, opened 2015-10-05
Finished assessment.

kaylal12:


lib/challenge.js
```diff
@@ -19,7 +19,24 @@
...
+// -manage events
+
+// 'Event'
+// Description: App manages events
+// Every 'event' has
+// -name of event
+// -time of event
+// -location
+// Every 'event' can
+// -alarm for reminder
```
GA-MEB, line 39: These are good abstractions for this problem - can you think of one more?

GA-MEB:
3.5 of 4 (missing a third abstraction for Q1)

Very nice work, Kayla.

PR #8, opened 2015-10-05
Committing response

atsundwn:


lib/challenge.js
```diff
@@ -19,6 +19,18 @@
...
+  // Every 'songe' has
+  // - Artist's name
+  // - Length of the song
+  // - Year of release
+  // - Genre
+  // Every 'song' can
+  // - be in a a playlist
+  // - play
+  // - stop
+  // - skipped
```
GA-MEB, line 33: I think you might need to re-read this question.

lib/challenge.js
```diff
@@ -51,7 +66,14 @@ var pet_owner;
 var some_pet;
 
 // Answer Starts Here
-
+some_pet = {
+  name: "Brownie",
+  species: "dog",
+  breed: "Border collie",
+  makeNoise: function(noise) {
+    console.log(noise);
```
GA-MEB, line 74: We were specifically looking for `makeNoise` to utilize the object's `noise` property; it should end up looking something like this:

``` javascript
somePet = {
  name : "Brownie",
  species : "dog",
  breed : "Border Collie",
  noise : "woof"
  makeNoise : function(){
    console.log(this.noise); // alternatively, "return this.noise;"
  } 
};
```

lib/challenge.js
```diff
@@ -60,7 +82,14 @@ var some_pet;
...
 // Answer Starts Here
-
+Pet = function Pet (name, species, breed, noise) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.makeNoise = function(noise) {
+    console.log(noise);
+  }
+}
```
GA-MEB, line 92: In this case, since our Constructor function will be making many identical objects, we don't want it to make a brand new copy of our makeNoise method each time. To avoid this, we can use the prototype of Pet, and define all of our methods there; since all individual pets will have access to Pet.prototype, they can still each use all those methods, but it will use a lot less memory.

``` javascript
  this.name = name;
  this.species = species;
  this.breed = breed;
  this.noise = noise;
}

  return this.noise;
};
```

GA-MEB:
2.5 of 4 (-0.5 on Q1, Q3, and Q4)

Let me know if you have any questions about any of this feedback!

PR #9, opened 2015-10-05
KMcMahon - added response to lib/challenge.js file

resource11:


GA-MEB:
It's all one pull request - a PR is just a flag on GitHub indicating that someone wants you to pull from their branch.

lib/challenge.js
```diff
@@ -19,7 +19,15 @@
   // - park
 
 // Answer Starts Here
-
+'Calendar'
+Description: our app uses a calendar to keep track of appointments.
+Every 'calendar' has
+- an appointment name
+- an appointment date
+- an appointment location  (ok, that is three, I tend to use calendars a bit).
```
GA-MEB, line 27: These sound like attributes of an "appointment" rather than a "calendar".

lib/challenge.js
```diff
@@ -19,7 +19,15 @@
...
+Every 'calendar' can
+- set busy/free status
+- invite another person to the appointment
```
GA-MEB, line 30: This abstraction isn't bad, especially if you make it an "appointment" instead of a "calendar", but it's only one. Can you think of any others?

lib/challenge.js
```diff
@@ -33,7 +41,12 @@
 var pet_owner;
 
 // Answer Starts Here
-
+var pet_owner = {
+  Owner : {
```
GA-MEB, line 45: This is a little redundant.

lib/challenge.js
resource11, line 30: At the moment, no. Because... I don't exactly understand the difference between an abstraction and an attribute. Much less so when being testing in a time-limited setting. :)

lib/challenge.js
resource11, line 45: Again, confused. How is it redundant?

lib/challenge.js
resource11, line 27: This is how I view a calendar. As an app that creates and tracks appointments. What am I missing here?

lib/challenge.js
```diff
@@ -51,7 +64,17 @@ var pet_owner;
...
-
+var some_pet = {
+  Pet : {
+    name : "Fluffy",
+    species : "cat",
+    breed : "Norwegian Forest",
+    noise : "meow",
+    makeNoise : function (noise) {
+      var uniqueNoise = noise;
+      return uniqueNoise;
```
GA-MEB, line 75: This should be referencing the `noise` attribute on `some_pet`, like so.

``` javascript
...
makeNoise : function() {
  console.log(this.noise);
},
...
```

lib/challenge.js
```diff
@@ -60,7 +83,16 @@ var some_pet;
 
 var Pet;
 // Answer Starts Here
-
+var Pet = function(name, species, breed, noise) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+  this.makeNoise = function(noise) {
```
GA-MEB, line 91: In this case, since our Constructor function will be making many identical objects, we don't want it to make a brand new copy of our makeNoise method each time. To avoid this, we can use the prototype of Pet, and define all of our methods there; since all individual pets will have access to Pet.prototype, they can still each use all those methods, but it will use a lot less memory.

``` javascript
  this.name = name;
  this.species = species;
  this.breed = breed;
  this.noise = noise;
}

  return this.noise;
};
```

GA-MEB:
2.5 of 4 (-0.5 on Q1, Q3, and Q4)

Let me know if you have any questions about this feedback, or about the topics in general.

PR #11, opened 2015-10-05
Challenge Complete

kelvinma:


lib/challenge.js
```diff
@@ -19,6 +19,19 @@
...
+  year:
+    leapYear: false
+  month
+    name: 'July',
+    season: 'Summer'
+    number of days: 30
+  day
+    date: 4,
+    dayOfWeek: 'Monday',
+    eventTime: '13:00:00',
```
GA-MEB, line 33: What does eventTime correspond to?

lib/challenge.js
```diff
@@ -58,9 +97,25 @@ var some_pet;
...
-var Pet;
 // Answer Starts Here
 
+var Pet = function(name, species, breed, noise){
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+  // option 1
+  this.makeNoise = function(){
```
GA-MEB, line 108: Nope - not this way. We don't want to define methods in our Constructor function, because that will cause multiple copies of our functions to be created. By putting methods on the prototype, the only get created once.

lib/challenge.js
```diff
@@ -58,9 +97,25 @@ var some_pet;
...
+    return this.noise;
+  }
+}
+
+// OR option 2
```
GA-MEB, line 113: Not 'or' - you need to do _both_. Attributes get defined in the Constructor, methods get defined on the Constructor's prototype.

GA-MEB:
3 of 4 (-0.5 on Q1, Q4)

Nice work. Let me know if you have any questions!

PR #12, opened 2015-10-05
completed assessment on objects intro

scottyscripts:
used no additional resources

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -19,7 +19,16 @@
   // - park
 
 // Answer Starts Here
-
+'Calendar'
+Description: This website acts like a calendar
+ 'calendar' has
+- list and layout of months days and years
+- space in each day to add events and meetings
+- list of holidays and birthdays
+the 'calendar ' can
+-send notifications about meetings and events
+-countdown days until specific events
+-synchronize with social media services
 // Answer Ends Here
 
 // 2. Create an Object literal that lines up with the following description. Store it in the variable 'pet_owner', below. Be sure to give it reasonable values for each of its properties.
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -51,7 +63,15 @@ var pet_owner;
...
 
 // Answer Starts Here
-
+some_pet = {
+  name : "Mr. Bigglesworth",
+  species : "cat",
+  breed : "sphinx",
+  noise : "meow",
+  makeNoise : function (noise) {
+    return noise;
```
GA-MEB, line 72: `return this.noise;`
Otherwise, we can't access the `noise` property.

lib/challenge.js
```diff
@@ -60,6 +80,16 @@ var some_pet;
...
 // Answer Starts Here
+var Pet = function Pet (name, species, breed, noise) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+}
+
+Pet.prototype.makeNoise = function makeNoise (noise) {
+  return noise;
```
GA-MEB, line 91: See above. Otherwise correct.

lib/challenge.js
```diff
@@ -19,7 +19,16 @@
...
+'Calendar'
+Description: This website acts like a calendar
+ 'calendar' has
+- list and layout of months days and years
+- space in each day to add events and meetings
+- list of holidays and birthdays
+the 'calendar ' can
+-send notifications about meetings and events
+-countdown days until specific events
+-synchronize with social media services
```
GA-MEB, line 31: Can you think of two other abstractions that might make sense?

GA-MEB:
2.5 of 4 (-0.5 on Q1, Q3, Q4)

Nice work; let me know if you have any questions.

PR #15, opened 2015-10-05
Filled out answers to challendge

mkutter72:


lib/challenge.js
```diff
@@ -19,6 +19,36 @@
...
+// Description:  represents a single day
+// Every Day has
+// - a name = day of week
+// - a list of appointments
+//
+//  Every day can
+// - Set a new appointment
+// - Delete an existing appointment
+
+
```
GA-MEB, line 50: Sure, not unreasonable! Some other abstractions we might consider for an application like this might be, for instance, 'Users', 'Events', or 'Calendars' (as in, individual calendars within the app).

lib/challenge.js
```diff
@@ -48,19 +81,37 @@ var pet_owner;
...
 
 // Answer Starts Here
 
+var some_pet = {
+    name : 'Sparky',
+    species : 'Dog',
+    breed : 'Pug',
+    noise: 'Bark',
+    makeNoise function(){
+      return noise;
```
GA-MEB, line 93: Should be `this.noise`, so that it can refer to the object's `noise` property.

GA-MEB:
3.5 of 4 (-0.5 on Q3)

Very nice work, Mike. Let me know if you have any questions!

PR #16, opened 2015-10-05
Responses to Object Assessments

baz1389:


lib/challenge.js
```diff
@@ -19,6 +19,15 @@
...
 // Answer Starts Here
+Every calendar has
+  - a year
+  - a month
+  - x amount of days in a month
+Every calendar can
+  - change year
+  -change month
+  -change day of the week
+
```
GA-MEB, line 30: Looking for another two abstractions - can you think of any?

lib/challenge.js
```diff
@@ -52,6 +66,13 @@ var some_pet;
 
 // Answer Starts Here
 
+some_pet = {
+  name : "Rowlf",
+  species : "Dog",
+  breed : "Cocker Spaniel",
+  noise : "Woof!"
+};
```
GA-MEB, line 74: Missing a `makeNoise` method, e.g. 

``` javascript
makeNoise: function(){
  return this.noise;
}
```

lib/challenge.js
```diff
@@ -60,6 +81,16 @@ var some_pet;
...
 var Pet;
 // Answer Starts Here
+Pet = function Pet(name, species, breed, noise){
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+};
+
+Pet.prototype.makeNoise = function makeNoise () { //I think I am misunderstanding why we need a function here if the pet noise is already a unique attribute that we defined in Q3
```
GA-MEB, line 91: No, you got it right!

`noise` is the property; `makeNoise` is the method that refers to it. And when you're using a constructor function, methods for new objects go into that constructor function's prototype.

GA-MEB:
3 of 4 (-0.5 on Q1, Q3)

Nice work - let me know if you have any questions!

PR #17, opened 2015-10-05
challenge

faetea:
ran out of time with the last question.

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -19,7 +19,16 @@
   // - park
 
 // Answer Starts Here
-
+'webCal'
+Description: An app to keep track of things via your browser.
+Every 'webCal' has (abstraction):
+  - login
+  - dates (year-month-day)
+  - color codes
+Every 'webCal' can (properties/behaviors):
+  - create/edit event
+  - delete event
+  - invite others to events
 // Answer Ends Here
 
 // 2. Create an Object literal that lines up with the following description. Store it in the variable 'pet_owner', below. Be sure to give it reasonable values for each of its properties.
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -19,7 +19,16 @@
...
+'webCal'
+Description: An app to keep track of things via your browser.
+Every 'webCal' has (abstraction):
+  - login
+  - dates (year-month-day)
+  - color codes
+Every 'webCal' can (properties/behaviors):
+  - create/edit event
+  - delete event
+  - invite others to events
```
GA-MEB, line 31: This is a good abstraction for this particular problem - can you think of two others?

lib/challenge.js
```diff
@@ -51,7 +63,12 @@ var pet_owner;
 var some_pet;
 
 // Answer Starts Here
-
+some_pet = {
+  name : "sophie",
+  species : "feline",
+  breed : "calico",
+  noise : "meow"
```
GA-MEB, line 70: Missing a `makeNoise` method, a la

``` javascript
makeNoise : function(){
  console.log(this.noise);
} 
```

lib/challenge.js
```diff
@@ -61,6 +78,17 @@ var some_pet;
...
+Pet = function Pet (name, address) {
+  this.name = name;
+  this.address = address;
+  this.some_pet = [];
+}
+
+Pet.prototype.specifics = function specifics () {
+  var name = some_pet.name
+  return name;
+};
```
GA-MEB, line 90: Not quite. The method should be `makeNoise`, and it should reference `this.noise`, like so:

``` javascript
  console.log(this.noise);
};
```

GA-MEB:
2.5 of 4 (-0.5 on Q1, Q3, Q4)

Nice work - let me know if you have any questions.

PR #18, opened 2015-10-05
response

drewster317:


lib/challenge.js
```diff
@@ -19,6 +19,16 @@
...
+ //"calendar"
+// Description: our program keeps track of daily, weekly, monthly things.
+// Every "calendar" has
+// - a day
+ //- a week
+ //- a month
+ //Every "calendar" can
+// - tell what the current day is
+ //- what holidays are in the month
+ //- show tasks by date
```
GA-MEB, line 31: This is a reasonable abstraction, but we were expecting three. Can you think of two others that would make sense for this application?

lib/challenge.js
```diff
@@ -30,9 +40,14 @@
...
   // - an address
 
-var pet_owner;
+var pet_owner = {
 
 // Answer Starts Here
+var owner = {
+Description : " Making an appointment at Newton Vet for cat check-up",
+name: "Drew McMullen",
+address: 250 Waltham Street, Newton, MA
```
GA-MEB, line 49: This should be a string.

lib/challenge.js
```diff
@@ -48,9 +63,17 @@ var pet_owner;
   // Every pet can:
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
-var some_pet;
+var some_pet {
 
 // Answer Starts Here
+Pet: cat,
```
GA-MEB, line 69: This will look for a variable `cat`, which I don't think is what you meant. If you wanted to have the value be a String, it needs to be in quotes (" "). The only time when Strings don't need quotes is when they are _keys_ on an object - keys don't need quotes.

lib/challenge.js
```diff
@@ -48,9 +63,17 @@ var pet_owner;
...
+description: "Making an appointment for shots",
+name: Obie,
+species: cat,
+breed: siamese,
+noise: meow,
+var makeNoise = "Meow" ;
```
GA-MEB, line 75: This isn't how you write methods. Methods are just like any other property on an object, except that their values are functions.

``` javascript
makeNoise : function(){
  console.log(this.noise);
}
```

GA-MEB:
1 of 4 (-0.5 on Q1, -0.5 on Q2, -1 on Q3, -1 on Q4)

You're not too far off, but it seems like you're having trouble with the syntax for objects. I think it might be beneficial for you to practice writing some objects in repl.it; if they're syntactically correct, you shouldn't hit any errors when you run the code.

PR #20, opened 2015-10-05
assessment response

sarabrentmccoy:


lib/challenge.js
```diff
@@ -20,6 +20,40 @@
...
+// Description: Our app can set off an alarm for no specified event, only at a specified time
+//Every 'alarm' has
+// - a tone
+// - a time
+// - a snooze option (yes or no)
+// Every 'alarm' can
+// - sound
+// - snooze
+// - repeat (every day, every Monday, etc.)
+
```
GA-MEB, line 54: Very nice! These are excellent abstractions.

lib/challenge.js
```diff
@@ -52,6 +91,16 @@ var some_pet;
 
 // Answer Starts Here
 
+var some_pet = {
+  name: 'Claire',
+  species: 'dog',
+  breed: 'Rottweiler',
+  noise: 'Bark',
+  makeNoise: f() {
+    return 'Bark!'
```
GA-MEB, line 100: Mmm.... 
We were looking for you to refer to the `noise` property here; i.e. `return this.noise`.

lib/challenge.js
```diff
@@ -61,6 +110,17 @@ var some_pet;
...
 
+var Pet = {
+  name: 'Claire',
+  species: 'dog',
+  breed: 'Rottweiler',
+  noise: 'Bark',
+
+makeNoise: function (){
+  return noise;
+}
```
GA-MEB, line 121: No, we don't want to do this because it would create a brand new function (i.e. Function object) every time we ran the constructor. That's wasteful, and could consume a lot of memory; it would be better to have one place to keep all our methods, so that they don't get created each time.

That's where `prototype` comes in. A Constructor's `.prototype` property is an object where you can keep all methods that you want instances (i.e. the output of your Constructor) to be able to access, without each having their own individual copy.

It's written like this:

``` javascript
  return this.noise;
};
```

GA-MEB:
3 of 4 (-0.5 on Q3, Q4)

Nice work - let me know if you have any questions.

PR #21, opened 2015-10-05
added answers to the challenge.js doc

christinemoses:


lib/challenge.js
```diff
@@ -34,6 +46,11 @@ var pet_owner;
 
 // Answer Starts Here
 
+var pet_owner = {
+  name: 'Christine',
+  address: '79 Chestnut St, Wilmington, MA 01887',
```
srhoulam, line 51: This is a good object literal.

The interpreter won't complain about that final comma, but recall that the comma is a separator character for key-value pairs. Think of it like how we type spaces to separate words in a sentence. Following this analogy, the closing curly brace is like a period, and our sentence is the object literal, starting with `{` and ending with `}`. Putting a comma after the final key-value pair, then, is like putting a space between the final word and the period that ends the sentence .

lib/challenge.js
```diff
@@ -52,6 +69,19 @@ var some_pet;
 
 // Answer Starts Here
 
+var some_pets = () {
```
srhoulam, line 72: That set of parens (`()`) has no place in this statement.

lib/challenge.js
```diff
@@ -52,6 +69,19 @@ var some_pet;
...
+  name: 'DiDi',
+  species: 'dog',
+  breed: 'labrador/beagle mix',
+  noise: [
+    {
+      happyNoise: 'bark',
+      angryNoise: 'growl',
+    }
+  ],
```
srhoulam, line 81: What you've assigned to the `noise` attribute here is a list of one element. It doesn't look like there's any purpose for additional elements, so it raises the question of why it's a list.

The question was looking for you to place a method on the object by assigning a function to the property `makeNoise`. Since what you've got between the curly braces is mostly good, I'll give this partial credit.

lib/challenge.js
```diff
@@ -61,6 +91,24 @@ var some_pet;
 var Pet;
 // Answer Starts Here
 
+var Pet = function Pet(name, species, breed, noise) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = []];
```
srhoulam, line 98: That second closing square bracket is a syntax error. Unmatched brackets are always trouble!

lib/challenge.js
```diff
@@ -61,6 +91,24 @@ var some_pet;
...
+
+};
```
srhoulam, line 100: Aside from that loose bracket, this is a perfectly good constructor function. Well done!

lib/challenge.js
```diff
@@ -61,6 +91,24 @@ var some_pet;
...
+
+Pet.prototype.makeNoise = function makeNoise () {
```
srhoulam, line 102: Awesome. This question was asking you to create a constructor function and attach a method to its prototype, which you did.

lib/challenge.js
```diff
@@ -61,6 +91,24 @@ var some_pet;
...
+  var makeNoises = 0;
```
srhoulam, line 103: This choice of variable name is very dangerous. It's a one-character difference from the name of the function!

In general, you do not give variables verb names unless they contain a function. A better choice would have been `noisesMade`.

lib/challenge.js
```diff
@@ -61,6 +91,24 @@ var some_pet;
...
+  for (var i = 0; i < this.noise.length; i++) {
+    makeNoises += this.noise[i].noise;
+  }
+  return makeNoise;
```
srhoulam, line 107: Here, that variable name I pointed out earlier bites you. This function doesn't return the variable `makeNoises`, but the function itself, which internally calls itself `makeNoise`!

lib/challenge.js
```diff
@@ -61,6 +91,24 @@ var some_pet;
...
+};
```
srhoulam, line 108: So, given that you chose to take the approach of iterating over an array of noises, and concatenating strings stored in the elements of that array, the initial value of your accumulator (the variable you're combining everything in) is contrary to your desires.

You set the initial value of `makeNoises` to `0`, a number. When you add a number to a string, the number gets turned into a string to make sense of it. As a result, your string of noises would start with `'0'`, e.g., `'0woofwoof'`.

A good initial value for a string accumulator is the empty string, `''`. It's like the zero of strings. If you concatenate it with another string, you get the other string every time, just like how if you add zero to another number, you get the other number every time.

lib/challenge.js
```diff
@@ -20,6 +20,18 @@
...
+every calendar has
+  - hours
+  - days
+  - months
+  - years
+every calendar can
+  - schedule appointments/events
+  - list to-do activities
+  - set an alert to notify you of an upcoming event
+
```
srhoulam, line 34: Well done, this is what was asked for. The question asked for 3 of these, though, so I can only give you half credit.

srhoulam:
2.5/4

PR #22, opened 2015-10-05
finished assessment

pguzman26:


lib/challenge.js
```diff
@@ -30,10 +40,14 @@
   // - a name
   // - an address
 
-var pet_owner;
 
-// Answer Starts Here
 
+// Answer Starts Here
+var pet_owner {
```
srhoulam, line 46: You need an assignment operator (`=`) here between `pet_owner` and `{`. As the code here stands, it will throw a syntax error because the interpreter doesn't know what to do with an object literal immediately following a variable declaration.

lib/challenge.js
```diff
@@ -30,10 +40,14 @@
...
+  clinic:
```
srhoulam, line 47: Since no value is supplied for this object literal's `clinic` attribute, this is an incomplete key-value pair. The missing comma, which serves to separate key-value pairs, means the interpreter combines the two lines into a syntactically invalid key-value pair: `clinic: owner: [name, address]`. An error will then be thrown when the second colon is read, since a key-value pair is allowed only one colon.

lib/challenge.js
```diff
@@ -30,10 +40,14 @@
...
+  owner: [name, address]
```
srhoulam, line 48: For an object representing a pet owner, who is, presumably, a human being, to have an attribute called `owner` suggests that our pet-owning human being has an owner, doesn't it?

A suggestion on lists (arrays): JS will not constrain you as to what types you place in lists, but practicality will. If you're placing values in a list, you usually want to be able to iterate over it, running the same code against every item in the list. For example, if you had a list of people to send a greeting card to, you could go through the list on by one and purchase a greeting card for each of them. Now, let's say you added an odd one out: "bake bread". You'd have to know that "bake bread" is not a name, and that you can't apply the same procedure for that value.

When building a list, build it with this in mind. You can iterate over a list of names, or a list of addresses, but a list of names and addresses would mean your loop logic would need to become more complex. In software development, you want to avoid complexity wherever possible.

lib/challenge.js
```diff
@@ -48,7 +62,14 @@ var pet_owner;
...
 
-var some_pet;
+var some_pet {
+
+name: ,
+species: ,
+breed: ,
+noise: ,
+
+ ;}
```
srhoulam, line 72: This is a mostly-good framework for an object literal. The colons are where they should be, and the commas are where they should be, save the last one -- the last key-value pair in an object literal should not be followed with a comma, no more than the last word in a sentence should be followed with a space.

The biggest problem here is that semicolon. Semicolons are not allowed characters in object literal syntax. They always lead to a syntax error. It should come after the closing curly brace.

A less serious problem is the absence of values for these four attributes. This is why I said it's a mostly-good framework; it's a good place to be before deciding which values to assign to each attribute. Scaffolding object literals like this can be very helpful.

lib/challenge.js
```diff
@@ -48,7 +62,14 @@ var pet_owner;
   // Every pet can:
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
-var some_pet;
+var some_pet {
```
srhoulam, line 65: You're missing the assignment operator (`=`) between `some_pet` and `{`. Are you experiencing any confusion with curly braces? They have several meanings depending on context. If you're unsure, come see me or someone else on the instructional team for clarification.

lib/challenge.js
```diff
@@ -58,9 +79,16 @@ var some_pet;
...
 //    which behaves as described in Q3.
 
-var Pet;
+var Pet = function Pet (name,species,breed,noise) {
+  this.name = name;
+  this.species = species;
+  this. breed = breed;
+  this.noise = noise;
+
+}
```
srhoulam, line 88: What you've got here is a perfectly good constructor function. Well done! This question was looking for you to create a method on `Pet`'s prototype so that all instances of `Pet` would have the method through it. Half credit.

srhoulam:
1.25/4

PR #24, opened 2015-10-05
response

LaurenGrant:


lib/challenge.js
```diff
@@ -33,6 +44,13 @@
 var pet_owner;
 
 // Answer Starts Here
+var pet_owner = {
+  name: Lauren Grant,
+  address: Somerville,
+  pet vaccinations: [
```
srhoulam, line 50: Recall how object property names are strings. In object literal syntax, you're allowed to just write the property names without quotation marks, like so:

``` js
var obj = {
    hasCookies : true
};
```

... but only so long as the property name consists entirely of valid name characters. This basically means "so long as it consists of nothing but consecutive letters". So, suppose we wanted to create an object literal with a  property named "favorite color". This would **not** work:

``` js
var obj = {
    favorite color : "blue"
};
```

The interpreter would see that space between "favorite" and "color" and start looking for a colon (`:`). It won't find it, and it will conclude that you're using incorrect syntax. Here's how we reclaim these potential property names:

``` js
var obj = {
    "favorite color" : "blue",
    "maybe-not" : false,
    "cookies+milk" : "yep",
    "!in&huh?" : null
};
```

In these cases, we have to wrap the property names in quotation marks. Once we do that, we're free to use whatever funny symbols we want, including spaces, hyphens, plus signs, and things we probably wouldn't even want to put in property names.

lib/challenge.js
```diff
@@ -33,6 +44,13 @@
...
+
+  ]
+}
```
srhoulam, line 53: Other than that, good object literal. 3/4 credit.

lib/challenge.js
```diff
@@ -48,7 +66,14 @@ var pet_owner;
   // Every pet can:
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
-var some_pet;
+var some_pet = {
+  name: "Spud",
+  species: "Dog",
+  breed: "Lab",
+  noise: = function () {
```
srhoulam, line 73: That equal sign is a syntax error in an object literal. However, if you simply remove just the equal sign, this is exactly what the question was asking for.

lib/challenge.js
```diff
@@ -48,7 +66,14 @@ var pet_owner;
...
+    return "I'm a" + species + "I say Woof Woof.";
+}
```
srhoulam, line 75: You never want to see two consecutive closing curly braces at the same level of indentation like this.This one should be indented one step, to show that it matches the opening curly brace from the function. The second one should have the same level of indentation as `var some_pet = {`, which is none here. This makes it easy to know which opening curly braces they match, and saves you reading time.

lib/challenge.js
```diff
@@ -48,7 +66,14 @@ var pet_owner;
...
+}
```
srhoulam, line 76: Well done. That `=` that needs to be removed is something you could easily figure out by trying to run this code and looking closely at it, and your answer mostly satisfies the question. 3/4 credit.

lib/challenge.js
```diff
@@ -58,7 +83,17 @@ var some_pet;
 //    the values of `name`, `species`, `breed`, and `noise`. The constructor function should also give each pet a method called `makeNoise`,
 //    which behaves as described in Q3.
 
-var Pet;
+var Pet = function(name, species, breed, noise) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+}
```
srhoulam, line 91: This is a good constructor function you've made here and assigned to `Pet`. We wanted you to add a prototype method on `Pet` so all of its instances will have it. Half credit.

srhoulam:
3/4

PR #25, opened 2015-10-05
functions 

francescadal:
i relied to heavily on looking up solutions online when i didn't understand what needed to be done.

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -19,6 +19,11 @@
   // - park
 
 // Answer Starts Here
+var car = {
+  make: 'Jeep',
+  model: 'Wrangler',
+  year: 2016,
+};
 
 // Answer Ends Here
 
@@ -30,7 +35,11 @@
   // - a name
   // - an address
 
-var pet_owner;
+var pet_owner = {
+  name:'baxter',
+  address: 'change st, somerville, ma 02145',;
+
+}
 
 // Answer Starts Here
 
@@ -48,7 +57,13 @@ var pet_owner;
   // Every pet can:
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
-var some_pet;
+var some_pet = {
+  name:'Baxter',
+  species: 'dog',
+  breed:'Golden Retriver'
+  noise: 'bark',
+  makeNoise:
+}
 
 // Answer Starts Here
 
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -19,6 +19,11 @@
   // - park
 
 // Answer Starts Here
+var car = {
+  make: 'Jeep',
+  model: 'Wrangler',
+  year: 2016,
+};
```
srhoulam, line 26: This is not what the question asked for, but you are assigning a mostly good object literal to `car`. That comma after `2016` is not necessary, since `year: 2016` is the final key-value pair in the object literal. The interpreter won't complain, though.

lib/challenge.js
```diff
@@ -30,7 +35,11 @@
   // - a name
   // - an address
 
-var pet_owner;
+var pet_owner = {
+  name:'baxter',
+  address: 'change st, somerville, ma 02145',;
```
srhoulam, line 40: This object literal is broken by that semicolon. Semicolons have no place in key-value pairs, which are all that is allowed directly inside an object literal. When a semicolon occurs here, then, the interpreter throws a syntax error, and does not create the object.

lib/challenge.js
```diff
@@ -30,7 +35,11 @@
...
+
+}
```
srhoulam, line 42: Other than the unnecessary final comma and the incorrect use of semicolon, this is a good object literal. Half credit.

lib/challenge.js
```diff
@@ -48,7 +57,13 @@ var pet_owner;
   // Every pet can:
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
-var some_pet;
+var some_pet = {
+  name:'Baxter',
+  species: 'dog',
+  breed:'Golden Retriver'
```
srhoulam, line 63: There is no comma on this line, which is required to separate key-value pairs. This object literal, when parsed, will cause the interpreter to throw an error between `'Golden Retriver'` and `noise:`.

lib/challenge.js
```diff
@@ -48,7 +57,13 @@ var pet_owner;
...
+  noise: 'bark',
+  makeNoise:
```
srhoulam, line 65: The value corresponding to the key `makeNoise` is expected to be a function. Properties/attributes of objects that point to functions are called methods.

lib/challenge.js
```diff
@@ -48,7 +57,13 @@ var pet_owner;
...
+}
```
srhoulam, line 66: Aside from not having the requested method and the missing comma, this is a well-formed object literal. Half credit.

lib/challenge.js
```diff
@@ -58,7 +73,15 @@ var some_pet;
...
 
-var Pet;
+var Pet = function Pet (name, species, breed, noise) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+  this.makeNoise = [];
+
+}
```
srhoulam, line 83: This is a good constructor function. The created object's (the instance's) `makeNoise` property should be a function, not an array. Actually we wanted you to give `Pet` instances a `makeNoise` method without putting a property on each object, by putting a prototype method on `Pet`. Half credit.

srhoulam:
1.8/4

PR #26, opened 2015-10-05
finished assesment

LaurisaNeuwirth:


lib/challenge.js
```diff
@@ -34,6 +45,13 @@ var pet_owner;
 
 // Answer Starts Here
 
+// Description: We are making an app for a veterinary clinic - it manages pets' vaccinations.
+var pet_owner = {
+
+  name: "Laurisa",
+  address: "193 Jackson Street"
+  {
```
srhoulam, line 53: This is an opening brace, not a closing brace. This will lead to a syntax error. However, you don't make this mistake anywhere else, so it's probably just a typo.

lib/challenge.js
```diff
@@ -52,6 +70,15 @@ var some_pet;
 
 // Answer Starts Here
 
+var some_pet = {
+
+  name: "Bailey Boxer",
+  species: "Boxer",
+  breed: "mixed",
+  noise: "Grrrr"
+  ],
```
srhoulam, line 79: This loose, unmatched closing square bracket is a syntax error. Commas are also not needed following the final key-value pair in an object literal.

lib/challenge.js
```diff
@@ -52,6 +70,15 @@ var some_pet;
...
+},
```
srhoulam, line 80: Aside from that loose square  and unnecessary comma bracket above, this is a good object literal. The question asks you to put a `makeNoise` method on it, though. Half credit.

lib/challenge.js
```diff
@@ -61,6 +88,17 @@ var some_pet;
...
 
+var Pet = function(name, species, breed, noise) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+  this.makeNoise = function(){
+    return "This pet makes a " + this.noise + ", noise.";
+  };
+}
```
srhoulam, line 99: This is a good constructor function, and the `makeNoise` method is correct. However, this question asks you to create a prototype method. A little over half credit.

srhoulam:
3.1/4

PR #27, opened 2015-10-05
Answered Questions

HoldenMills:


lib/challenge.js
```diff
@@ -19,7 +19,16 @@
   // - park
 
 // Answer Starts Here
+var computer {
```
srhoulam, line 22: You're missing an operator here (between `computer` and `{`) for the assignment. Without an assignment operator, this statement is not well-formed.

lib/challenge.js
```diff
@@ -19,7 +19,16 @@
...
+  // Description our app managers computer characteristics//
+  manufacture: "Dell"
+  speed: '1.96'
+  memory: '900'
+  sendEmail: ["recipient", "address"]
+  togglePower: ["on", "off"]
+  moveMouse: ["left", "right", "up", "down"]
 
+}
```
srhoulam, line 31: This is a good object literal, but for one thing: if an object literal has multiple properties, the key-value pairs (e.g., `propertyName : 'a string value'`) must be separated by commas. Otherwise, the interpreter will not know where one ends and the next begins.

lib/challenge.js
```diff
@@ -33,7 +42,10 @@
 var pet_owner;
 
 // Answer Starts Here
-
+ {
+ name: "Jane Doe"
+ address: "1 Toll Rd."
+ }
```
srhoulam, line 48: The same missing commas apply here. This object literal is not being assigned at all, though. That means the reference is immediately lost upon creation.

lib/challenge.js
```diff
@@ -51,7 +63,14 @@ var pet_owner;
 var some_pet;
 
 // Answer Starts Here
-
+  name: "Sir Reginald"
```
srhoulam, line 66: The curly braces for an object literal are missing here.

lib/challenge.js
```diff
@@ -51,7 +63,14 @@ var pet_owner;
...
+  species: "Canine"
+  breed: "Corgi"
+  makeNoise : [
```
srhoulam, line 69: The expression assigned to the `makeNoise` property here is an array containing a single object literal, not a function. The object literal is not well-formed, though.

lib/challenge.js
```diff
@@ -60,7 +79,11 @@ var some_pet;
 
 var Pet;
 // Answer Starts Here
-
+var Pet = function(name, species, breed, nosie)
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = makeNoise;
```
srhoulam, line 86: This constructor function is well-formed except for the missing curly braces. The function body (where the statements that get run are) must be surrounded by curly braces. The assignments here are good, and usage of parameters is mostly good -- there is no `makeNoise` parameter taken by this function, but that looks more like a typo than anything. Finally, the requested prototype method is missing.

srhoulam:
1/4

PR #28, opened 2015-10-05
Adding my answers

gracenjambi:
I used the notes from class when answering the questions.

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -20,6 +20,17 @@
 
 // Answer Starts Here
 
+//'User'
+//Description our calendar manages tasks
+//Every date should have
+//- start date and time
+//- duration
+//- end date and time
+//Every task will have
+// - task description
+// - duration
+// - recurrence
+
 // Answer Ends Here
 
 // 2. Create an Object literal that lines up with the following description. Store it in the variable 'pet_owner', below. Be sure to give it reasonable values for each of its properties.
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -20,6 +20,17 @@
...
+//Description our calendar manages tasks
+//Every date should have
+//- start date and time
+//- duration
+//- end date and time
+//Every task will have
+// - task description
+// - duration
+// - recurrence
+
```
srhoulam, line 33: You broke the format here, describing the properties of dates and tasks, whereas the question would have been satisfied with you dropping one of those, but asked for a list of calendar behaviors, as well. Half credit.

lib/challenge.js
```diff
@@ -33,6 +44,11 @@
 var pet_owner;
 
 // Answer Starts Here
+pet_owner (date,vaccination){
+  date: "10-10-15"
+  vaccination: "Rabies vaccination"
+  }
```
srhoulam, line 50: This is a mix of function literal notation and object literal notation. Partial credit.

lib/challenge.js
```diff
@@ -51,6 +67,14 @@ var pet_owner;
...
 
 // Answer Starts Here
+some_pet(name,species,breed,noise){
+  [
+name: oreo
+species: dog
+breed:Rottweiler
+noise: bark
+  ]
+}
```
srhoulam, line 77: This is like above, but with the addition of square (array) brackets. Things it seems you intended as strings (`oreo`, `dog`, etc.) are actually references to variables that don't exist because the required quotation marks are missing.

srhoulam:
.75/4

PR #30, opened 2015-10-05
Add responses

bbarbersox:


lib/challenge.js
```diff
@@ -19,6 +19,36 @@
...
+// Every 'Calender' has
+// - months
+// - weeks
+//  - days
+// - dates
+// - events
+// Every calender can
+// - display schedule
+// - add new events
+// - modify events
```
srhoulam, line 32: Well done. The question asked for three or more, so this counts for half credit.

lib/challenge.js
```diff
@@ -30,10 +60,24 @@
...
+  address: 'Boston',
+  pets: {
+    name: 'Brutus',
+    type: 'dog',
+    vaccinations: {
+      vacName: rabies,
+      vacDate: '12-14-2104'
+    },
+  },
+},
```
srhoulam, line 77: Very good. You went beyond what the question asked for. For `pets`, if it's going to be plural, it should be able to contain more than one pet. An array containing objects similar to the object literal you built would be perfect.

lib/challenge.js
```diff
@@ -48,10 +92,17 @@ var pet_owner;
...
-
 // Answer Starts Here
 
+var some_pet = {
+  name: 'maggie',
+  species: 'cat',
+  breed: 'tiger''
+  noise: function() {
+    console.log('meow');
+};
```
srhoulam, line 103: The curly brace on this line closes the function body; the object literal remains unclosed. The semicolon, then, is a syntax error. The semicolon on the line above is not, though. Can you explain why? Partial credit.

lib/challenge.js
```diff
@@ -60,6 +111,12 @@ var some_pet;
 
 var Pet;
 // Answer Starts Here
+var Pet = function(name, species, breed, noise){
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+}
```
srhoulam, line 119: So far, so good. We were looking for you to put a prototype method on Pet that uses `this` to reference the object and access its `noise` property. Partial credit.

srhoulam:
3/4

PR #36, opened 2015-10-25
Response

resource11:
Reworked the my responses on lib/challenge.js 

Please see inline comments as well. Thanks!

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -88,7 +88,7 @@ var Pet = function(name, species, breed, noise) {
   this.species = species;
   this.breed = breed;
   this.noise = noise;
-  makeNoise = function(noise) {
+  this.makeNoise = function(noise) {
     var this.uniqueNoise = noise;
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -19,6 +19,31 @@
...
+- an hour
+- a day
+- a month
+- a year
+
+Each 'appointment' has
+- has a 'block of time assignment'
+- a 'name'
+- a 'location'
+
```
srhoulam, line 46: This is plain English in a code file. It should all be commented. Because these are not comments, you can't use the tests that would help you catch minor errors before submitting the assessment -- the script will blow up at the very top from syntax errors.

srhoulam:
Kathleen, please get the tests working and resubmit when you've got them all passing. `grunt nag` will give you best-practices suggestions, and `grunt test` will tell you exactly how your answers satisfy (or don't) the questions.

PR #38, opened 2015-10-25
Shimelassessment

eshimel:
Reworked some of these and it makes a lot more sense...

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -19,7 +19,17 @@
   // - park
 
 // Answer Starts Here
-
+// 'Calendar'
+//Description: Our calendar keeps track of things.
+// Every 'Calendar' has
+// - the months of the year
+// - days of the week
+// - weekly to-do list
+//Every 'Calendar' can
+// - navigate between dates
+// - add items to the to-do list
+// - send reminders
+//
 // Answer Ends Here
 
 // 2. Create an Object literal that lines up with the following description. Store it in the variable 'pet_owner', below. Be sure to give it reasonable values for each of its properties.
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -19,8 +19,34 @@
...
+- toggle between dates
+
+Every 'Calendar' has
+-days of the week
+- a daily to-do list
+- a monthly to-do list
+Every 'Calendar' can
+-toggle between dates
+-generate to-do lists
+-send reminders// Answer Ends Here
```
srhoulam, line 49: Great. The question asks for three such big-picture models, and you've got three.

One thing to note is the answer to this question is plain English, and this is a file containing code. Plain English belongs strictly in comments so that it doesn't prevent the file from being run as code. As a result, running the tests always results in a complete failure no matter how correct your answers are -- there are syntax errors at the very top of the file.

1

lib/challenge.js
```diff
@@ -30,10 +56,14 @@
...
-var pet_owner;
 
-// Answer Starts Here
 
+// Answer Starts Here
+var pet_owner = {
+  ownerFirstName: "someone's firstName",
+  LastName: "someone's last name",
+  Address: "address"
+};
```
srhoulam, line 66: Great. The only thing there is to note here is naming conventions. We don't name variables or properties with a capital initial unless they're constructor functions.

1

srhoulam:
Edie, since your answer the first question prevents using the tests, I recommend use `grunt nag` and `grunt test` to further work on this assessment and resubmit in the future.

PR #39, opened 2015-10-26
Response

pguzman26:
updated

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -7,6 +7,7 @@
 // Write your answers inside this file, at the places where it's indicated by the comments.
 
 // 1. Suppose that we wanted to create a browser-based calendar program for keeping track of things. In comments, list at least three reasonable abstractions that you might use to build this program; for each abstraction, list out at least two properties and behaviors that it would make sense for that abstraction to have. Below is an example - please write your answer in the same format.
+
   // 'Car'
   // Description: Our app manages car sales.
   // Every 'car' has
@@ -20,6 +21,15 @@
 
 // Answer Starts Here
 
+// Abstractions would be for 'Calendar'
+// Calendar would have
+// - Year
+// - Month
+// - Day
+
+// Each 'Day' will have
+// The time of thing being tracked
+// Then I would add maybe coloring certains tasks that are being tracked
 // Answer Ends Here
 
 // 2. Create an Object literal that lines up with the following description. Store it in the variable 'pet_owner', below. Be sure to give it reasonable values for each of its properties.
@@ -30,10 +40,14 @@
   // - a name
   // - an address
 
-var pet_owner;
+
 
 // Answer Starts Here
-
+var pet_owner {
+  clinic:
+  owner: [name, address]
+}
+;
 // Answer Ends Here
 
 // 3. Create an Object literal that lines up with the following description. Store it in the variable `some_pet`, below.
@@ -48,7 +62,14 @@ var pet_owner;
   // Every pet can:
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
-var some_pet;
+var some_pet {
+
+name: ,
+species: ,
+breed: ,
+noise: ,
+
+ ;}
 
 // Answer Starts Here
 
@@ -58,8 +79,15 @@ var some_pet;
 //    the values of `name`, `species`, `breed`, and `noise`. The constructor function should also give each pet a method called `makeNoise`,
 //    which behaves as described in Q3.
 
-var Pet;
+var Pet = function Pet (name,species,breed,noise) {
+  this.name = name;
+  this.species = species;
+  this. breed = breed;
+  this.noise = noise;
+
+}
 // Answer Starts Here
+
 
 // Answer Ends Here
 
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -20,6 +21,15 @@
...
 
+// Abstractions would be for 'Calendar'
+// Calendar would have
+// - Year
+// - Month
+// - Day
+
+// Each 'Day' will have
+// The time of thing being tracked
+// Then I would add maybe coloring certains tasks that are being tracked
```
srhoulam, line 32: Great. The question asks for three such descriptions, though. 1/2

lib/challenge.js
```diff
@@ -30,10 +40,14 @@
...
-var pet_owner;
 
-// Answer Starts Here
 
+// Answer Starts Here
+var pet_owner {
+  Owner: "John Doe",
+  name: "Max",
+  address:"21 franklin st."
+}
```
srhoulam, line 50: `var variableName` declares a new variable. It is itself a complete statement. The variable begins with a value of `undefined`.

``` js
{
    prop1: val1,
    prop2: val2
}
```

The above is an object literal. It is a complete value expression.

Following the variable declaration with a value expression, however, is grammatically incorrect. It's like following a noun with another noun in English. There has to be something else in between that describes how they relate, or what to do with one or both of them.

What is missing here (and in your next answer) is the **assignment operator (`=`)**. You want to assign the object literal to the variable `pet_owner`. You achieve this by putting the assignment operator between the variable name and the value expression, like so:

``` js
var pet_owner = {
  Owner: "John Doe",
  name: "Max",
  address:"21 franklin st."
};
```

3/4

lib/challenge.js
```diff
@@ -48,7 +62,19 @@ var pet_owner;
   // Every pet can:
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
-var some_pet;
+var some_pet {
+
+name: "John Doe",
+species: "Dob",
+breed: "Chia",
+noise: [
```
srhoulam, line 70: This property contains a reference to an array that contains one object literal. The actual noise strings are stored in the object literal's properties. The array seems to serve no purpose here.

lib/challenge.js
```diff
@@ -48,7 +62,19 @@ var pet_owner;
   // Every pet can:
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
-var some_pet;
+var some_pet {
```
srhoulam, line 65: Missing an assignment operator here.

lib/challenge.js
```diff
@@ -48,7 +62,19 @@ var pet_owner;
...
+species: "Dob",
+breed: "Chia",
+noise: [
+{
+  noise1: "woof",
+  noise2: "WOOOOOF",
+  noise3: "meow"
+}
+],
+ ;}
```
srhoulam, line 77: Semicolons have no place inside an object literal. Without fail, they will result in a syntax error being thrown. There are two types of curly braces: the kind the follow functions, loops, and `if`s, which contain **statements**, and the kind that contain **key-value pairs**.

Statements are actions. They lead to things being done. The semicolon (`;`) is the statement terminator. It's like a period at the end of a sentence.

Key-value pairs are essentially property assignments. Their syntax is as follows:

`keyName: value`

That is, an identifier (or string), a colon (`:`), and a value expression. Multiple key-value pairs must be separated using commas (`,`). While directly inside an object literal, only key-value pairs are syntactically correct.

Back to the question, it wanted you to place a method on an object literal to log a pet-like sound. This object literal doesn't have a method. Here is an example solution that could potentially help:

``` js
var some_pet = {
    name : 'Whiskers',
    species : 'Weasel',
    breed : null, // do weasels have breeds?
    noise : 'Weasel sound!',
    makeNoise : function() {
        console.log(this.noise);
    }
};
```

1/2

lib/challenge.js
```diff
@@ -58,8 +84,15 @@ var some_pet;
...
 //    which behaves as described in Q3.
 
-var Pet;
+var Pet = function Pet (name,species,breed,noise) {
+  this.name = name;
+  this.species = species;
+  this. breed = breed;
+  this.noise = noise;
+
+}
```
srhoulam, line 93: This is a good constructor function and should work flawlessly.

lib/challenge.js
```diff
@@ -58,8 +84,15 @@ var some_pet;
...
 // Answer Starts Here
+var somePet = new Pet("max", "dog"."lab","woof!");
```
srhoulam, line 95: You've got a period between `"dog"` and `"lab"` where there should be a comma. It's a typo, so it won't be counted, but you should know in order to be more careful in the future.

This is otherwise a good instantiation (creating a specific pet using the `Pet` constructor).

What this question wanted in addition to the constructor function was for you to place a prototype method on it that all instances of `Pet` would then have. Here is an example of putting a method on a prototype:

``` js
    console.log(this.noise);
};
```

1/2

srhoulam:
2.25/4

jrhorn424:
@gmankd Please use `grunt nag` and `grunt test` to check your code for correctness. `grunt nag` would have caught your assignment mistakes. `grunt test` will tell you whether a question is correct before you move on.

PR #40, opened 2015-10-27
Response

faetea:
resubmitting with corrections to questions I got wrong

### DIFF
--- a/lib/challenge.js
+++ b/lib/challenge.js
@@ -61,13 +61,15 @@ pet_owner = {
   // - make noise (each pet makes its own unique noise, as specified by `noise`.
 
 var some_pet;
-
 // Answer Starts Here
 some_pet = {
   name : "sophie",
   species : "feline",
   breed : "calico",
   noise : "meow"
+  makeNoise : function(){
+    console.log(this.noise);
+  }
 }
 // Answer Ends Here
 
@@ -77,16 +79,14 @@ some_pet = {
 
 var Pet;
 // Answer Starts Here
-
-Pet = function Pet (name, address) {
+Pet = function Pet (name, species, breed, noise) {
   this.name = name;
-  this.address = address;
-  this.some_pet = [];
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
 }
-
-Pet.prototype.specifics = function specifics () {
-  var name = some_pet.name
-  return name;
+Pet.prototype.makeNoise = function() {
+  console.log(this.noise);
 };
 
 // Answer Ends Here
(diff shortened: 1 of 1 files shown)


lib/challenge.js
```diff
@@ -19,7 +19,16 @@
...
+'webCal'
+Description: An app to keep track of things via your browser.
+Every 'webCal' has (abstraction):
+  - login
+  - dates (year-month-day)
+  - color codes
+Every 'webCal' can (properties/behaviors):
+  - create/edit event
+  - delete event
+  - invite others to events
```
srhoulam, line 31: This is good modeling. Trouble is that this is plain-English text in a code file and it's not commented out. As a result, you were unable to run the included tests on your potential solutions, which would have caught and helped you find errors that would keep your code from running.

This question asked for three such models, and here you've got one. 1/2

lib/challenge.js
```diff
@@ -33,7 +42,10 @@
 var pet_owner;
 
 // Answer Starts Here
-
+pet_owner = {
+  name : "Saphron Reyn",
+  address : "1 Second St, Cambridge, MA, 02141",
+}
```
srhoulam, line 48: Your object literal is good in that it will run correctly. Run `grunt nag` if you'd like to further polish it; there are some ways it could be better.

1

lib/challenge.js
```diff
@@ -49,9 +61,16 @@ var pet_owner;
...
-
+some_pet = {
+  name : "sophie",
+  species : "feline",
+  breed : "calico",
+  noise : "meow"
+  makeNoise : function(){
+    console.log(this.noise);
+  }
+}
```
srhoulam, line 73: This object literal is syntactically incorrect; the tests (run `grunt test`) would have caught this, but not explained why, and `grunt nag` and your linter will show you exactly what's wrong here.

Bottom line, this is one minor fix from being entirely correct. 1

lib/challenge.js
```diff
@@ -60,6 +79,15 @@ var some_pet;
 
 var Pet;
 // Answer Starts Here
+Pet = function Pet (name, species, breed, noise) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.noise = noise;
+}
```
srhoulam, line 87: Good constructor function.

lib/challenge.js
```diff
@@ -60,6 +79,15 @@ var some_pet;
...
+Pet.prototype.makeNoise = function() {
+  console.log(this.noise);
+};
```
srhoulam, line 90: Good prototype method. Well done. 1

srhoulam:
3.5/4

PR #41, opened 2015-10-27
Re-submission

HoldenMills:


lib/challenge.js
```diff
@@ -20,6 +20,14 @@
 
 // Answer Starts Here
 
+var alarm clock = {
+    display: 'digital',
+    color: 'black',
+    weight: '8 ounces',
+    togglePowerSource: ['AC', 'battery']
```
srhoulam, line 27: You need commas at the end of this line and at the end of the line that follows it. Commas separate key-value pairs; the array after the colon is the value segment of this key-value pair.

lib/challenge.js
```diff
@@ -20,6 +20,14 @@
...
+    togglePower: ['on', 'off']
+    setAlarm: ['time', 'noise']
+}
```
srhoulam, line 30: This is not at all what the question was asking for, but your enthusiasm for writing object literals should be rewarded. 1/2

lib/challenge.js
```diff
@@ -33,7 +41,10 @@
 var pet_owner;
 
 // Answer Starts Here
-
+var pet_owner = {
+  name: 'Jake, Elwood',
+  address: "1060 West Addison, Chicago, IL"
+}
```
srhoulam, line 47: Great. 1

lib/challenge.js
```diff
@@ -51,7 +62,12 @@ var pet_owner;
 var some_pet;
 
 // Answer Starts Here
-
+var some_pet = {
+  name: 'Holly',
+  species: "Canis lupis Familiarus",
+  breed: 'Corgi',
+  makeNoise: function(bark)
```
srhoulam, line 69: Very much on the right track. The function in the value part of this key-value pair is missing a body, though. If it had a body, you could put statements in the body, and could complete the question exactly.

lib/challenge.js
```diff
@@ -51,7 +62,12 @@ var pet_owner;
...
+}
```
srhoulam, line 70: 4/5

lib/challenge.js
```diff
@@ -60,7 +76,11 @@ var some_pet;
 
 var Pet;
 // Answer Starts Here
-
+var Pet = function(name, species, breed, noise)
```
srhoulam, line 79: This function is missing the curly braces that denote its body. How would anyone be able to tell which statements belong to the function, and which ones are outside of it?

lib/challenge.js
```diff
@@ -60,7 +76,11 @@ var some_pet;
...
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.makeNoise = function(bark)
```
srhoulam, line 83: This question wanted you to put the method on the prototype, not the instance.

Also, I'm noticing a recurring lack of curly braces following a function expression. Here is the **minimal function**, a function lacking everything but what is necessary to be syntactically correct: `function() {}`

Curly braces must occur even if the function has no statements to go inside them. They're non-negotiable. They're how the interpreter knows which statements are within the function, and if they're empty, they're how the interpreter knows there are no statements within the function. If they're absent, the interpreter concludes the code is broken and gives up.

Here's a sample solution to this question:

``` js
var Pet = function Pet(name, species, breed, noise) {
    this.name = name;
    this.species = species;
    this.breed = breed;
    this.noise = noise;
}
    console.log(this.noise);
};
```

2/5

srhoulam:
2.7/5

PR #49, opened 2016-01-22
js-objects-diagnostic

apalmer0:


lib/diagnostic.js
```diff
@@ -17,7 +17,13 @@
...
 
 // Answer Starts Here
-//
+Entity: User
+Attributes: name
+Methods: createEvent, deleteEvent
+
+Entity: Event
+Attributes: description, startTime, location, duration
+Methods: alert, remind, repeat
```
GA-MEB, line 26: Good choices for these. As a thought experiment, see if you can figure out how you might actually implement a `repeat` method!

lib/diagnostic.js
```diff
@@ -38,19 +48,33 @@ let owner = null;
...
-let pet = null;
+let pet = {
+      name: 'Skunk',
+      species: 'dog',
+      breed: 'golden retreiver',
+      vocalization: 'woof',
+      vocalize: function(){
+        return 'my name is ' + name + ' and i say ' + vocalization;
+      }
+    };
```
GA-MEB, line 59: Not quite - the `name` and `vocalization` properties aren't visible from inside the `vocalize` function; can you think of a way to reference them?

lib/diagnostic.js
```diff
@@ -38,19 +48,33 @@ let owner = null;
...
 // `species`, `breed`, and `vocalization`.
 
-const Pet = function() {
-  /* Your answer here */
+const Pet = function(name, species, breed, vocalization) {
+    this.name = name,
+    this.species = species,
+    this.breed = breed,
+    this.vocalization = vocalization
+    }
```
GA-MEB, line 70: Two `}`s here - this would cause your code not to run.

GA-MEB, line 70: Otherwise correct.

lib/diagnostic.js
```diff
@@ -38,19 +48,33 @@ let owner = null;
...
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+Pet.prototype.vocalize = function(){
+  return 'wooooooooof';
```
GA-MEB, line 76: As in Q3, this method should print out the value stored on a pet's `vocalize` property. Can you think of how you might do that?

GA-MEB:
3.25 / 5 (-0.5 on Q3, -0.25 on Q4, -1 on Q5)

PR #52, opened 2016-01-22
HERE ARE MY ANSWERS

kpie89:
I hope they're okay.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,19 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+Description: Calender program for keeping track of things
+Entity: Event
+Attributes: when, where, duration
+Methods: conflictingEvent
+
+Entity: User
+Attributes: name, address, e-mail
+Methods: schedule, outOfTown
+
+Entity: Date
+Attributes: day, month, year
+
+
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following description.
@@ -28,7 +40,11 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
+let owner = {
+  name: 'Kevin';
+  address: '31 Chester Street'
+  numberOfPets: 4;
+}
 
 // 3. Create an object literal that lines up with the following description.
 // Store it in the variable `pet`, below.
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -17,7 +17,19 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+Description: Calender program for keeping track of things
+Entity: Event
+Attributes: when, where, duration
+Methods: conflictingEvent
```
GA-MEB, line 23: Just so that you think about it, `conflictingEvent` might need to take another Event as an argument to determine if they overlap. Also, it would probably be more conventional to call it `isConflictingEvent(blah)`, since it will presumably be returning a boolean.

lib/diagnostic.js
```diff
@@ -17,7 +17,19 @@
...
+
+Entity: User
+Attributes: name, address, e-mail
+Methods: schedule, outOfTown
```
GA-MEB, line 27: `outOfTown` sounds like a property (and, as above, it should probably be called `isOutOfTown` since it'd be giving us a boolean). Why do you think it makes sense for it to be a method? Is it performing some kind of calculation to determine if the user is out of town?

GA-MEB, line 27: `schedule` is also a little ambiguously named. Do you mean "schedule a new event"? "Show the current schedule"?

lib/diagnostic.js
```diff
@@ -38,19 +54,36 @@ let owner = null;
...
-let pet = null;
+let pet = {
+  name: 'Bubbles';
+  species: 'Dog';
+  breed: 'Lab';
+  vocalization: function() {
+    this.vocalize
+  }
+  return vocalize;
+}
```
GA-MEB, line 65: Not quite. `vocalization` was supposed to be a property representing the noise that the pet makes (e.g. 'woof'), while `vocalize` was supposed to be a method that returned (or printed - the directions could be interpreted either way) the value of `vocalization`.

lib/diagnostic.js
```diff
@@ -38,19 +54,36 @@ let owner = null;
...
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+Pet.prototype.vocalize = function() {
+  vocalization: function() {
+    this.vocalize
+  }
+  return vocalize;
+}
+}
```
GA-MEB, line 86: You correctly added a method to the prototype (`Pet.prototype.methodName = function(){}`), but the function you added is wrong. See my notes on Q3, above.

GA-MEB:
3.5 / 5 (-0.5 on Q1, -0.5 on Q3, -0.5 on Q5)

Generally looks good. Let me know if you have any questions about my feedback.

PR #53, opened 2016-01-22
Byau

byau:


lib/diagnostic.js
```diff
@@ -17,7 +17,12 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+// Entity: workcalendar
+// Attributes: time, dates, event name, invitees, list
+// Methods: invite, decline, add event, delete event, change time/dates, add/delete items in list, add/delete list
+// Entity: personalcalendar
+// Attributes: time, dates, event name, list
+// Methods: add event, delete event, change time/dates, add/delete items in list, add/delete list
```
GA-MEB, line 25: It seems like `workCalendar` is the same as `personalCalendar`, but with some additional bells and whistles. When it comes time to actually code this, you might want `workCalendar` to _inherit_ from `personalCalendar`. 

[Inheritance](https://en.wikipedia.org/wiki/Inheritance_%28object-oriented_programming%29) is a way to define objects based on other objects that have already been defined. JavaScript has a fairly unusual inheritance pattern (prototypal), which we'll learn more about in Unit 3. However, we'll see another kind of inheritance (classical) much sooner when we learn Ruby in Unit 2.

GA-MEB:
5 / 5

Nice work.

PR #63, opened 2016-01-22
response to diagnostic

msyao:
Sorry it's late - had problem

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -16,8 +16,16 @@
 // Attributes: name, commissionRate
 // Methods: sell, call
 
-// Answer Starts Here
-//
+   Answer
+   Description: Keep track of things in calendar.
+   Entity: Calendar
+   Attributes: day, week, month, year
+   Methods: add data, alert,track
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -28,7 +36,12 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
+
+let Owner = null {
```
RealWeeks, line 40: I think you may have meant to remove null?

RealWeeks:
4/5.  I don't think you understand prototypes.  

Here is why:

 +const Pet = function (){
-  this.name = name;
-  this.species = species;
-  this.vocalization = vocalization;
-  this.vocalize = [];
  +};

That is creating a constructor function.  You want something like this:
- Pet.prototype.vocalize = function(){
  -    return this.vocalization;
    +};

I recommend you review the repos from that day, maybe read the docs.  Your code is generally good so I'm not that worried but let me know if you want a bit more explanation.

PR #67, opened 2016-03-25
Add response to js-objects-diagnostic

cuprous:
And maybe just a bit of funny?


RDegnen:
5 out of 5 fantastic!

PR #69, opened 2016-03-25
Responds to 'js-objects-diagnostic'

OllieBuilds:


RDegnen:
5 out of 5, but you can fill all those values in they don't have to be empty strings

PR #73, opened 2016-03-25
Diagnostic, Mar. 25

susandaiute:
Answer questions on objects

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -16,9 +16,17 @@
 // Attributes: name, commissionRate
 // Methods: sell, call
 
-// Answer Starts Here
-//
-// Answer Ends Here
+Description: browser-based calendar program that keeps track of the calendar for
+an entire family.
+Entity1: familyMember
+Attributes: name, nickname, phone, email
+Methods: use nickName, call, email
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -28,7 +36,11 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
+let owner = {
+  name: "Jenny",
+  address: "Whatever bar is closest",
+  numberOfPets: 17
```
RDegnen, line 42: It seems Jenny has a recurring problem

lib/diagnostic.js
```diff
@@ -38,19 +50,29 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: "Cheetoh",
+  species: "Cat",
+  breed: "Shorthair",
+  vocalization; "Meow"
```
RDegnen, line 57: You forgot to add the method here

RDegnen:
4.5 out of 5 cause you just missed one tiny detail

PR #74, opened 2016-03-25
adds answers

V4Vince:


lib/diagnostic.js
```diff
@@ -38,19 +46,33 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: "Mr.Paws",
+  species: "dog",
+  breed: "Siberian Husky",
+  vocalize: function(vocalization){
+    let vocalization = "bark bark"
```
RDegnen, line 54: you are declaring vocalization as variable here AND passing it in as an argument which will break the code

lib/diagnostic.js
```diff
@@ -38,19 +46,33 @@ let owner = null;
...
+  this.breed = breed,
+  this.vocalization = vocalization,
+  this.vocalize = vocalize
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+const vocalize = function(){
+  return this.vocalization,
```
RDegnen, line 74: You ended this line with a `,` which throws an `SyntaxError: Unexpected token }` and this.vocalization is undefined here. aaand you did not declare this function as a prototype of Pet

lib/diagnostic.js
```diff
@@ -38,19 +46,33 @@ let owner = null;
...
 // `species`, `breed`, and `vocalization`.
 
-const Pet = function() {
-  /* Your answer here */
+const Pet = function(name, species, breed, vocalization) {
+  this.name = name,
+  this.species = species,
+  this.breed = breed,
+  this.vocalization = vocalization,
+  this.vocalize = vocalize
```
RDegnen, line 68: `this.vocalize = vocalize' will not work because you have not vocalize is not passed in as an argument

RDegnen:
2 out of 5

PR #75, opened 2016-03-25
Complete js objects diagnostic

MichaelB167:


lib/diagnostic.js
```diff
@@ -38,19 +50,32 @@ let owner = null;
...
 
 // Replace null with your answer
-let pet = null;
+let Pet = {
+  name: "Roger",
+  species: "Dog",
+  breed: "Corgie",
+  vocalization: "bark",
+  vocalize: function() {
+    console.log(Pet.vocalization)
```
RealWeeks, line 59: This works, but we want to return not console.log usually

RealWeeks:
5/5. Really nice work Michael.  Great job.  Keep up the effort and keep having the growth mindset.

PR #90, opened 2016-05-20
Object prototype diagnostics

shireen-bean:
Creating object literals, constructors and adding to prototypes
Ran grunt test but it doesnt like my Kobe vocalization even if I have it say return this._vocalization.
Also I wanted to make the animal say the noise so I'm just leaving it the way it is

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,10 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+//Description: Out program manages fitness and health done on each day
+//Entity: Person
+//Attributes: Height, weight, age
+//Methods: input new workout, calculate calories burnt, caclulate water drank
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -38,18 +45,33 @@ let owner = null;
...
-let pet = null;
+let pet = {
+  name: "Kobe",
+  species: "dog",
+  breed: "Japanese Spitz",
+  vocalization: "Yap! Yap!",
+  vocalize: function (){
+    console.log(pet.vocalization);
+  }
+};
```
RDegnen, line 56: you want to RETURN a string, the test wont pass for a console.log so 

``` js
vocalize: function (){
  return this.vocalization;
}
```

RDegnen:
4 of 5 nice

PR #91, opened 2016-05-20
JS Objects Diagnostic

andreasantarl:
Created constructors and methods for pet object

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,16 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+//Description: Calendar that keeps track of things
+
+//Entity #1: One day of the year
+// Attributes: Day, month, year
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -38,19 +51,33 @@ let owner = null;
...
-let pet = null;
+let pet = {
+    name: 'Percy',
+    species: 'Cat',
+    breed: 'Chunky monkey',
+    _vocalization: 'meow',
+
+    vocalize: function (){
+      console.log (this.name + ' says ' + this._vocalization + ' when he wants a butt scratch.');
+    };
```
RDegnen, line 62: instead of `console.log` you should `return` that string. Also the semi colon at the end of the function is a syntax error, because this function is a property of the pet object. So you could replace that semi colon with nothing or a `,`

RDegnen:
4 of 5 not bad!

PR #92, opened 2016-05-20
objects diagnostic

jacobhereford:
finished the objects diagnostic and answered all questions

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,6 +17,11 @@
 // Methods: sell, call
 
 // Answer Starts Here
+//description: My app is a recipe book
+// entity: recipes
+// attributes: which ingredients, how much for each ingredients,
+//              what kind of pan/pot needed
+// methods: prep time, cooking time, total time,
 //
 // Answer Ends Here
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -28,7 +33,13 @@
...
 
 // Replace null with your answer
-let owner = null;
+let owner = [
+  name: Jacob Hereford
+  age: 30,
+  address: 123 nonya
+  city: Boston
+  numberOfPets: 2
+];
```
RDegnen, line 42: Your defining this with `[ ]` instead of `{ }` which makes it an array. Also you don't have any comm's seperating each property.

lib/diagnostic.js
```diff
@@ -38,19 +49,32 @@ let owner = null;
...
 
 // Replace null with your answer
-let pet = null;
+let pet = [
+  name: bubbles
+  species: cat
+  breed: feral
+  vocalization: meow
+  vocalize: reow
+];
```
RDegnen, line 58: Same as above

lib/diagnostic.js
```diff
@@ -38,19 +49,32 @@ let owner = null;
...
 
-const Pet = function() {
-  /* Your answer here */
+const Pet = function(name, species, breed, vocalization, vocalize) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.vocalization = vocalization;
+  this._vocalize = vocalize;
+
```
RDegnen, line 70: Good, but you dont need the `_vocalize` property

RDegnen:
2 of 5

PR #99, opened 2016-05-20
Diagnostics Esther js-objects

esthercordova:


lib/diagnostic.js
```diff
@@ -45,19 +45,37 @@ let owner = {
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: "Charly",
+  species: "dog",
+  breed: "mixed",
+  vocalization: "bark bark"
+};
```
RDegnen, line 53: This is where you were supposed to put the vocalize method.

``` js
vocalize: function() {
  return this.vocalization;
}
```

lib/diagnostic.js
```diff
@@ -45,19 +45,37 @@ let owner = {
...
+
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+Pet.prototype.vocalize = function () {
+  return this._vocalization;
+};
+
+//to use the vocalize method type: cat.vocalize();
+// the vocalize method is not working- why? Could you please explain? Thanks.
```
RDegnen, line 78: This is correct

RDegnen:
4 of 5 good job

PR #103, opened 2016-07-15
David Powell Diagnostic

dvdpwll:
Filled in answers to questions and passed grunt test.


laurenfazah:
Great effort, David! All answers are correct.

PR #106, opened 2016-07-15
Answered questions

cdelsoin:
I should've gotten here on time!

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,13 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+let event = (host, name, date, time); {
+  this.host = host,
+  this.name = name,
+  this.date = date,
+  this.time = time,
+}
+
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity description.
@@ -28,8 +34,11 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
-
+let owner = (name, address, numberOfPets) {
+this.name = name,
+this.address = address,
+this.numberOfPets = numberOfPets
+}
 // 3. Create an object literal that lines up with the following `pet` entity description.
 // Store it in the variable `pet`, below.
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -28,8 +34,11 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
-
+let owner = (name, address, numberOfPets) {
```
laurenfazah, line 37: I think you were overthinking this. Something like the following would work:

```
let owner = {
  name: 'Brian',
  address: '41 West St.',
  numberOfPets: 2
};
```

lib/diagnostic.js
```diff
@@ -38,7 +47,12 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = (name, species, breed, vocalization) {
```
laurenfazah, line 50: Close! We don't need a constructor yet, though. Method name should be `vocalization` here.

``` js
let pet = {
  name: 'Beau',
  species: 'Dog',
  breed: 'Lab',
  vocalization: 'bark',
  vocalize: function () {
    return this.vocalization;
  }
};
```

lib/diagnostic.js
```diff
@@ -38,7 +47,12 @@ let owner = null;
...
+  this.name = name,
+  this.species = species,
+  this.breed = breed,
+  this.vocalization = vocalization(){}
+}
```
laurenfazah, line 56: Q4 and Q5 solutions:

``` js

const Pet = function(name, species, breed, vocalization) {
 this.name = name;
 this.species = species;
 this.breed = breed;
};

const vocalize = function () {
  return "bark"
};

Pet.prototype.vocalize = vocalize;
```

lib/diagnostic.js
```diff
@@ -17,7 +17,13 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+let event = (host, name, date, time); {
```
laurenfazah, line 20: This just needed to follow the same structure as the examples given above - the goal was to abstract properties of the calendar app.

laurenfazah:
Yeah, being on time definitely helps with finishing in time. 

No tests were passing for a number of reasons (both from syntax errors and misunderstanding the questions). Let me know if you'd like to go over any of these.

PR #136, opened 2016-10-15
Ci pass

BenGitsCode:
CI Test example of Passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: node_js
+node_js:
+- '0.10'
+before_install: gem install compass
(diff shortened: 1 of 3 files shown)


gaand:
I think you want to fix the warnings and not use force.

gaand:
Also, why is `gem install compass` necessary?

BenGitsCode:
I agree, and the gem isn't. I just wanted to see if I could get the CI status to show from the main pull request page, without opening each. I'm working on the error that I forced here in another branch that I'll make a PR for tonight.

PR #140, opened 2016-10-21
Finish js-objects-diagnostic

kiraamaa:
Use Google and class notes to respond

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,15 @@
 // Methods: sell, call
 
 // Answer Starts Here
+// Description: Our calendar keeps track of important dates and events.
 //
+// Entity: User
+// Attributes: name, email
+// Methods: createEvent, showConflict
+//
+// Entity: Events
+// Attributes: name, date, time, location
+// Methods: list, reminder
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity description.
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -17,7 +17,15 @@
 // Methods: sell, call
 
 // Answer Starts Here
+// Description: Our calendar keeps track of important dates and events.
 //
+// Entity: User
+// Attributes: name, email
+// Methods: createEvent, showConflict
```
neugierige, line 24: `createEvent` is certainly a method that users should have access to. But I don't think it should belong on the user, but rather the calendar. 

Similarly, are you sure `showConflict` should be a method on `User`? The ability to show two events with overlapping times would require knowledge about all the events. That seems to be a functionality of the calendar itself.

Conversely, one method that we might see on a user is `subscribed` - which stores information about whether the user is subscribed to our calendar. Does that make sense?

lib/diagnostic.js
```diff
@@ -38,19 +50,35 @@ let owner = null;
...
 // `species`, `breed`, and `vocalization`.
 
-const Pet = function () {
-  /* Your answer here */
+const Pet = function (name, species, breed, vocalization) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.vocalization = vocalization;
 };
```
neugierige: Very close!
```
let pet = {
  name: 'Beau',
  species: 'Dog',
  breed: 'Lab',
  vocalization: 'bark',
  vocalize: function () {
    return this.vocalization;
  }
};
```

neugierige:
4/5. Great job!

PR #145, opened 2016-10-21
js-objects-diagnostic response

treep78:


lib/diagnostic.js
```diff
@@ -28,7 +37,12 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
+let owner =
+{
+  name: 'name',
+  address: 'address',
+  numberOfPets: 'numberOfPets',
+};
```
BenGitsCode, line 46:
``` js
let owner = {
  name: 'Brian',
  address: '41 West St.',
  numberOfPets: 2
};
```

let's count it but the , after numberOfPets is not necessary

lib/diagnostic.js
```diff
@@ -38,19 +52,37 @@ let owner = null;
...
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+Pet.prototype.vocalize = function ()
+{
+  return this.vocalization;
+};
 
 // Don't change this
 module.exports = {
```
BenGitsCode, line 88:
``` js
const vocalize = function () {
  return "bark"
};

Pet.prototype.vocalize = vocalize;
```

BenGitsCode (commented):
4/5

PR #147, opened 2016-10-21
Attempted first of diagnostic

jbuzzanco:
tried to complete first question of diagnostic, am probably getting it wrong.

Comfort 1
clarity 2

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -16,10 +16,25 @@
 // Attributes: name, commissionRate
 // Methods: sell, call
 
-// Answer Starts Here
-//
-// Answer Ends Here
+const carSales = function (make, model, year, price, SalesPerson) {
+  this.make = "Audi";
+  this.model = "A7";
+  this.year = 2012;
+  this.car = true;
+  this.price = 45000;
+  this._salesPerson= function {
+    name = "Alexandria"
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -16,10 +16,25 @@
...
+    name = "Alexandria"
+    commissionRate = 4 * this.price;
+  };
+};
+
+
+
+  }
+
+  ]
```
BenGitsCode, line 36: You were overthinking this one I think.

``` js
// Write your answers inside this file, at the places where it's indicated by the comments.

// 1. Suppose that we wanted to create a browser-based calendar program for
// keeping track of things. Identify at least two entites such a program might
// need and describe attributes and methods of each.
//
// EXAMPLE
// Description: Our app manages car sales.
// Entity: Car
// Attributes: make, model, year
// Methods: drive, brake, park
//
// Entity: SalesPerson
// Attributes: name, commissionRate
// Methods: sell, call
```

think of it as a blueprint for beginning the example, a calendar app.

The example does the same with a car dealership so an entity of calendar might be reminders or monthview

BenGitsCode (commented):
1/5

Please review the solutions, and if you're still having trouble with them reach out to a consultant for a one on one to go over them.

PR #164, opened 2016-11-18
answer all questions, not sure if intrepeted what they wanted me to d…

RFSync:
…o correctly


JosephinaTavera:
🎧 

Quick notes: 

😂  chupacabra 😹 

Question 5, they were looking for you to use 'prototype' to attach the method and to return a value.

PR #168, opened 2016-11-18
response

cliftonh02:
Could have done better.


tvcrawley:
There were a few unanswered questions. If you check out the solution branch, you'll be able to see the type of answer the questions were looking for.

PR #174, opened 2016-11-18
All set with my answers

SandyPok:
I ran the grunt test and answered all of my questions.


tvcrawley:
👍  Good job. If you check out the solution branch, you'll be able to see the answers for the diagnostic and compare them to your own.

PR #175, opened 2016-11-18
Nov 18, 2016 Alyssa Felix

awanderlyss:


JosephinaTavera:
🎧 

Here is additional reading material:

http://javascriptissexy.com/javascript-prototype-in-plain-detailed-language/

PR #177, opened 2016-11-18
answers to diagnostic

patrickthesailorman:


JosephinaTavera:
🎧 

Here is a link that could help with understanding functions: 

http://www.javascriptkit.com/javatutors/primer6.shtml

PR #182, opened 2017-01-17
Respond to diagnostic

samordonez:
-added response to diagnostic

Comfort: 4
Clarity: 5

I used Google, and specifically a stack exchange question from that search, to check that I was properly declaring a function inside an object literal.


bernardlee:
5/5

PR #186, opened 2017-01-17
Maura's answers

ladypinetree:
comfort: 2-3
clarity: 3
post-weekend brain fog: 11

I used my notes and google

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -16,9 +16,14 @@
 // Attributes: name, commissionRate
 // Methods: sell, call
 
-// Answer Starts Here
-//
-// Answer Ends Here
+//Description: Our app organizes appointments and events
+//Entity: Meeting
+//Attributes: time, location, attendeeName
+// Methods: schedule, selectLocation
+
+//Entity: Organizer
+//Attributes: name, email, phoneNumber
+// Methods: create, invite, confirm
 
 // 2. Create an object literal that lines up with the following `owner` entity description.
 // Store it in the variable below. Be sure to give it reasonable values for each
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -28,7 +33,7 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
+let owner = ['name', 'address', 'numberOfPets'];
```
bernardlee, line 36: This creates an array with three elements, not an object with attributes.

lib/diagnostic.js
```diff
@@ -38,19 +43,28 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = ['name', 'species', 'breed', 'vocalization'];
```
bernardlee, line 46: This creates an array with four elements, not an object with attributes.

lib/diagnostic.js
```diff
@@ -38,19 +43,28 @@ let owner = null;
...
+  return this._vocalization;
+};
+//answer to #5
+
+const Pet = function (name, species, breed, vocalization) {
+  this.name = name;
+  this.species = species;
+  this. breed = breed;
+  this._vocalization= vocalization;
+  this.useVocalization = useVocalization;
```
bernardlee, line 62: This works, but in the future please don't rename named identifiers so that the tests work correctly.  e.g., this diagnostic asks for a method called ```vocalize``` not ```useVocalization```.

bernardlee:
3/5

PR #187, opened 2017-01-17
I wrote a function for the defined objects

novakw267:
Pet and owner.

I felt okay with this once I reminded myself what an object literal was. I had to look at the previous two lessons just to feel confident I understood it, but once I did i felt good. I ran out of time before i could fix what grunt test wanted me to fix though. 
I used this to refine my definition for what an object literal was. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types

comfort 3
clarity 3
I am feeling better about this, and functions, but definitely want to still go over it a little more.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,10 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+//description: browser-based calander
+//entity: calander
+//Attributes: month, day, year, display,
+//methods: leaving notes, pinning events, setting reminders
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity description.
@@ -28,7 +31,9 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
+  let owner = { name: 'jim',
+    address: '85 webster street',
+    numberOfpets: '2',};
 
 // 3. Create an object literal that lines up with the following `pet` entity description.
 // Store it in the variable `pet`, below.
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -38,19 +43,29 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+  let pet = {
+    animal: 'dog',
+    breed: 'corgi',
+    name: 'snugglemuffin',
+    vocalization: 'bork',
+};
```
bernardlee, line 51: Missing ```species``` attribute.

lib/diagnostic.js
```diff
@@ -17,7 +17,10 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+//description: browser-based calander
+//entity: calander
+//Attributes: month, day, year, display,
+//methods: leaving notes, pinning events, setting reminders
 // Answer Ends Here
```
bernardlee, line 24: Missing second entity example.

lib/diagnostic.js
```diff
@@ -38,19 +43,29 @@ let owner = null;
...
+  this.animal = animal;
+  this.breed = breed;
+  this.name = name;
+  this.vocalization = vocalization;
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+Pet.prototype.vocalization = function() {
```
bernardlee, line 66: This should be named ```vocalize```.

bernardlee:
2/5

PR #188, opened 2017-01-17
Create Objects, methods and prototypes

wilsonkaya:
Create objects, methods and prototype

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,13 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+
+let forCar = function(make, model, year){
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -17,7 +17,13 @@
...
 
 // Answer Starts Here
-//
+
+let forCar = function(make, model, year){
+  this.make = make;
+  this.model = model;
+  this.year = year;
+
+}
```
bernardlee, line 26: We want two examples of entities along with their attributes and methods for a browser-based calendar application.

bernardlee:
4/5

PR #190, opened 2017-01-17
Answers diagnostic questions

SquirtleSquad1988:
comfort: 4.5
clarity: 4.5 
I used the MDN javascript guide to help answer these questions

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,16 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+description: the app is a calendar that allows the user to create new entries
+each day as things to do. The app also allows individuals to set daily reminders
+entity: Daily reminder
+attributes: day of reminder. message itself. snooze or dismiss ability.
+Methods: notify user of alarm. create new reminder. delete reminder.
+
+entity: calendar itself.
+attributes: days, months, year. Holidays and phases of moon.
+Methods: change from one calendar day to another. change from one calendar months
+to another.
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity description.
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -28,7 +37,10 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
```
bernardlee, line 31: We're looking for something with the following form.
```javascript
let owner = {
  name: 'Someone',
  address: 'Some place',
  numberOfPets: 1
};
```
What you gave was a nested object with ```owner.Attributes``` as the object that we're looking for.

lib/diagnostic.js
```diff
@@ -38,19 +50,30 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
```
bernardlee, line 41: We're looking for something with the following form.
```javascript
const pet = {
  name: 'some name',
  species: 'some species',
  breed: 'some breed',
  vocalization: 'some sound'
};
```
What you gave was a nested object with ```pet.Attributes``` as the object that we're looking for.  An error is also thrown because ```vocalize``` doesn't exist.

lib/diagnostic.js
```diff
@@ -38,19 +50,30 @@ let owner = null;
...
+           Methods: vocalize};
+
 
 // 4. Create a constructor function for pet objects, as described above. The
 // constructor function should be responsible for setting the values of `name`,
 // `species`, `breed`, and `vocalization`.
 
-const Pet = function () {
-  /* Your answer here */
+const Pet = function (name, species, breed, vocaliation) {
```
bernardlee, line 65: ```vocaliation``` should be ```vocalization```.

lib/diagnostic.js
```diff
@@ -38,19 +50,30 @@ let owner = null;
...
+  this.species = species;
+  this.breed = breed;
+  this.vocalization = vocalization;
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+Pet.prototype.vocalize = function () {
+  return vocalization;
```
bernardlee, line 75: This should be ```return this.vocalization;``` otherwise it will throw an error since ```vocalization``` doesn't exist.

bernardlee:
2/5

PR #191, opened 2017-01-17
Submit Diagnostic

jwaskow:
Comfort: 4
Clarity: 4

I used my class notes and MDN to help with this diagnostic.  I had trouble with setting up the last prototype but I think I did everything else right.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,14 @@
 // Methods: sell, call
 
 // Answer Starts Here
+//Description: Our app keeps track of events on a calendar
+//Entity: calendar
+//Attributes: month, day, year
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -38,19 +49,33 @@ let owner = null;
...
+  this.species = species;
+  this. breed = breed;
+  this.vocalization = vocalization;
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
 /* your answer here */
+Pet.prototype.vocalize = function () {
+  console.log(this.vocalization);
```
bernardlee, line 77: This should be ```return this.vocalization;```.

bernardlee:
4/5

PR #192, opened 2017-01-17
diagnostic jan-17

ggabrakadabra:
-completed diagnostic questions

Comfort:3
Clarity:4

I checked on MDN for info about Methods and how/when to use them. This wasn't very helpful though because I'm still unsure how to use them correctly.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,6 +17,20 @@
 // Methods: sell, call
 
 // Answer Starts Here
+// Entity: User
+// Attributes: name, email,
+// Methods:
+//
+// Entity: Week 1
+// Attributes: wake up time, time it takes to get ready, time to go to work
+// Methods:
+//
+// Entity: Weekend 1
+// Attributes: time to go to birthday party, time to study
+// Methods:
+//
+// I understand that methods are what the entity 'does' but I'm not really
+// what to put here
 //
 // Answer Ends Here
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -38,20 +56,36 @@ let owner = null;
...
 
 // Replace null with your answer
-let pet = null;
-
+let pet = {
+  name: 'Majin Buu',
+  species: 'dog',
+  breed: 'mutt',
+  vocalization: 'bark',
+    let vocalize = function () {
```
bernardlee, line 64: You can't declare functions inside of objects like this.  The correct way is
```javascript
let pet = {
  name: 'some name',
  vocalize: function() { ... }
};
```

lib/diagnostic.js
```diff
@@ -38,20 +56,36 @@ let owner = null;
...
+// I know there are errors here, but methods are a little confusing to me still
+//
 // 4. Create a constructor function for pet objects, as described above. The
 // constructor function should be responsible for setting the values of `name`,
 // `species`, `breed`, and `vocalization`.
 
-const Pet = function () {
-  /* Your answer here */
+const Pet = function (name, breed, vocalization) {
+  name: 'name',
```
bernardlee, line 77: The ```species``` attribute is missing and these lines should be
```
this.name = name;
this.breed = breed;
```
The way that it's written right now looks like a mix of a function expression and an object literal and this will throw an error.

lib/diagnostic.js
```diff
@@ -38,20 +56,36 @@ let owner = null;
...
+  breed: 'breed',
+  vocalization: 'vocalization',
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+let Pet = function(name, breed, vocalize)
 
+Pet.prototype.vocalization(vocalize)
```
bernardlee, line 86:
```javascript
  return this.vocalization;
}
```

bernardlee:
2/5

PR #200, opened 2017-01-17
js-objects-diagnostic commit

cshea93:
comfort: 3
clarity: 4

A little confused with the abstraction/modeling.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -16,8 +16,15 @@
 // Attributes: name, commissionRate
 // Methods: sell, call
 
-// Answer Starts Here
+//ANSWER BELOW_____
+// Description: Our app is a calander system for keeping track of things
+//   Entity: Event
+//   Attributes: date, start time, end time
+//   Methods: Add event, remove event, edit event
 //
+//   Entity: User
+//   Attributes: name, date of birth, location, events
+//   Methods: event reminder
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity description.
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -38,19 +49,41 @@ let owner = null;
...
+  this.species = species;
+  this.breed = breed;
+  this.vocalization = vocalization;
 };
 
+
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
-/* your answer here */
+const Pet = function (name, species, breed, vocalization) {
```
bernardlee, line 76: ```Pet``` was declared earlier with ```const```.  This will throw an error since ```Pet``` is supposed to be a constant and it's being redeclared/redefined here.

lib/diagnostic.js
```diff
@@ -38,19 +49,41 @@ let owner = null;
...
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.vocalization = vocalization;
+};
+
+
+  pet.prototype.vocalize = function (){
```
bernardlee, line 84: ```pet``` should be ```Pet```.

lib/diagnostic.js
```diff
@@ -38,19 +49,41 @@ let owner = null;
...
+        return this.vocalize;
```
bernardlee, line 85: This should be ```return this.vocalization;```.  In its current state, it will just return the function itself.

bernardlee:
4/5

PR #201, opened 2017-01-17
Diagnostic response 1.17.2017

K0iram:
comfort: 4
clarity: 3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,6 +17,15 @@
 // Methods: sell, call
 
 // Answer Starts Here
+
+Entity: calendar
+Attributes: day, month, year
+Methods: findAppointment
+
+Entity: user
+Atributes: name, appointments
+Methods: createDate, deleteDate
+
 //
 // Answer Ends Here
 
@@ -28,7 +37,12 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
+let owner = {
+   name: Mario;
+   address: '555 Noname st';
+   numberOfPets: 4;
+}
+
 
 // 3. Create an object literal that lines up with the following `pet` entity description.
 // Store it in the variable `pet`, below.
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -28,7 +37,12 @@
 // Attributes: name, address, numberOfPets
 
 // Replace null with your answer
-let owner = null;
+let owner = {
+   name: Mario;
```
bernardlee, line 41: This should be ```name: 'Mario',```.  The way that it's currently written will cause this to throw an error because it will look for something that doesn't exist called Mario and also because `;` is not a valid delimiter for object properties.

lib/diagnostic.js
```diff
@@ -38,18 +52,34 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
```
bernardlee, line 55: The values for each key needs to be wrapped in quotation marks since they are supposed to be strings.  Right now, the interpreter will think that they're names of things that don't exist and will throw an error.

lib/diagnostic.js
```diff
@@ -38,18 +52,34 @@ let owner = null;
...
+  vocalization: meow;
+}
 
 // 4. Create a constructor function for pet objects, as described above. The
 // constructor function should be responsible for setting the values of `name`,
 // `species`, `breed`, and `vocalization`.
 
-const Pet = function () {
+const Pet = function (name, species, breed, vocalization) {
+  let name = this.name;
```
bernardlee, line 67: Almost.  The assignments are reversed.  They should be
```javascript
this.name = name;
this.species = species;
this.breed = breed;
```

lib/diagnostic.js
```diff
@@ -38,18 +52,34 @@ let owner = null;
...
+  let species = this.species;
+  let breed = this.breed;
+  let vocalization =this.vocalization;
   /* Your answer here */
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.
 
+Pet.prototype.vocalize = function () {
```
bernardlee, line 76: We just want to return the value of the object's ```vocalization``` attribute.
```javascript
  return this.vocalization;
}
```

bernardlee:
1/5

PR #203, opened 2017-01-17
Complete diagnostic.

lbreguet:
comfort: 3
clarity: 4
used notes on constructor functions and lab.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,36 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+const car = {
+  make: 'Toyota',
+  model: 'Prius',
+  year: 2016,
+  drive: function () {
+    let go = 0;
+    if (go < 1) {
+      return 'Drive';
+    }
+  },
+  brake: function () {
+    let stop = 1;
+    if (stop === 1) {
+      return 'Hold On';
+    }
+  },
+  park: function (){
+
+  },
+};
+const salesPerson = {
+  name: 'Louis Breguet',
+  commissionRate: 0.15,
+  sell: function (price) {
+    return price * this.commissionRate;
+  },
+  call: function () {
+
+  },
+};
 // Answer Ends Here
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -17,7 +17,36 @@
...
+const salesPerson = {
+  name: 'Louis Breguet',
+  commissionRate: 0.15,
+  sell: function (price) {
+    return price * this.commissionRate;
+  },
+  call: function () {
+
+  },
+};
```
bernardlee, line 49: We're just looking for two examples of entities, along with their attributes and methods, that can be used in a browser-based calendar application.

lib/diagnostic.js
```diff
@@ -38,16 +71,36 @@ let owner = null;
...
 
 // 4. Create a constructor function for pet objects, as described above. The
 // constructor function should be responsible for setting the values of `name`,
 // `species`, `breed`, and `vocalization`.
 
-const Pet = function () {
-  /* Your answer here */
+const Pet = function (species, breed, vocalization) {
```
bernardlee, line 93: Missing ```name``` in the parameter list.

lib/diagnostic.js
```diff
@@ -38,16 +71,36 @@ let owner = null;
...
+  this.species = species;
```
bernardlee, line 94: Missing assignment of ```this.name``` to ```name```.

lib/diagnostic.js
```diff
@@ -38,16 +71,36 @@ let owner = null;
...
+  this.breed = breed;
+  this.vocalization = vocalization;
 };
 
+const vocalization = function () {
+
+}
```
bernardlee, line 101: We want ```vocalize()``` to return the value of the object's ```vocalization``` attribute and to be available to all objects of the type ```Pet```.
```javascript
  return this.vocalization;
};
```

bernardlee:
2/5

PR #218, opened 2017-03-27
Add response to js-objects diagnostic

jim-moody:
Comfort: 5
Clarity: 5

Mostly went off what I remembered. I had to look at the notes for prototypes to remember if we can use "this" inside the method we add to a prototype.


benjimelito:
Everything looks good here, nice job. 5/5

PR #219, opened 2017-03-27
Responses - JS Objects Diagnostic

naributler:
Comfort: 4
Clarity: 4

I used js-objects-prototypes/lib/hero.js to complete this diagnostic. 
I did have trouble running grunt test. I'm not sure how to run the test in terminal. Could someone advise on this?

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,16 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+Entity: Event Information
+  Attributes: Subject, location, date, time
+
+Entity: Recurrence Information
+  Attributes: Day of recurrence, time of recurrence
+  Methods: setRecurrenceFrequency
+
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,20 +52,33 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: 'Jyo',
+  species: 'cat',
+  breed: 'domestic shorthair',
+  vocalization: 'meow',
+  vocalize: function () {}
```
benjimelito, line 60: You want to make sure your method returns something, in this case the inside should look something like this: `return this.vocalization`

benjimelito:
Looks good! I'm not sure the issue you're having running `grunt test` though, what error is it giving you? I would make sure that you're running `npm install` before messing with the repo, but that's just a guess. 4/5

PR #221, opened 2017-03-27
Submitting Response

sjruby:
Clarity 5/5
Comfort 4/5

I used class notes for the constructor sections, I feel preety comfortable
with functions/object but probably need to practice more on constructors.
I know I need to use THIS keywords inside of constructor functions but
don't really understand WHY I need to do it!  Will be doing some rereading
the this function to figure that out.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,13 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+// Enity: user
+// Attributes: email, user name, assignedTasks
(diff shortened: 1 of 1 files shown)


benjimelito:
Nice work! Wrapping your head around `this` can be a difficult concept to master, but I would say that to answer your question, in this case we are using `this` so that our function is re-usable in multiple scenarios. For example, imagine you want to make 2 `Cat` objects, one that used 'meow' as its vocalization, and one that used 'purr'. By using `this` in your `vocalization` method, you can reference two different vocalizations without having to rewrite them inside the function itself. 5/5

PR #222, opened 2017-03-27
Added response to diagnostic.md

ghost:


ghost:
Comfort: 4
Clarity: 4

I used Google and my class notes to help with this diagnostic - needed reminders on syntax.

lib/diagnostic.js
```diff
@@ -39,20 +52,39 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+
+let pet = {
+  name: 'doggy',
+  species: 'canine',
+  breed: 'labradoodle',
+  vocalize: function () {
```
benjimelito, line 60: You want this function to return a property on the `pet` object, rather than returning a variable that you define inside the function itself. For example, you would want to include a property called vocalization: `vocalization: 'woouuaafuh'` and then have your `bark` method look like this: `vocalize: function(){return this.vocalization}`

benjimelito:
Looks good, but make sure that your method is returning a property on your object (for this exercise). 4/5

PR #223, opened 2017-03-27
Modify diagnostic.js for js-objects-diagnostic

jkjellan:
I added answers in diagnostic.js, creating object literals,
as well as constructors and prototype methods.

comfort: 5
clarity: 5


benjimelito:
Good stuff 🌈  5/5

PR #226, opened 2017-03-27
JS Objects Diagnostic

andyjzhong:
Comfort: 4 / 5
Clarity: 3.9 / 5

Used notes from class and notes from self studying this weekend.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,22 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+
+Description: Our calendar helps Danny Ainge figure out whether he needs to make a trade at the trade deadline.
+Entity: Boston Celtics, Danny Ainge
+Attributes: roster, coach,
+Methods: trade, release, sign
+
+Entity: DannyAinge
+Attributes: name, dateOfBirth, hometown, salary
+Methods: trade, release, sign
+
+Entity: BostonCeltics
+Attributes: teamName, hometown
+Methods: makePlayoffs
+
+// Either the directions were not specific enough or my creativity in the morning is null. :(
+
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,20 +58,33 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: 'Octavia',
+  species: 'Gecko',
+  breed: 'Leopard',
+  vocalization: 'Moooooo'
+};
```
benjimelito, line 66: Here you want to include a method that will return the `vocalization` you have added. Something like this: `vocalize: function() {return this.vocalization}`

lib/diagnostic.js
```diff
@@ -39,20 +58,33 @@ let owner = null;
...
+  this.species = species
+  this.breed = breed
+  this._vocalization = speak
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.  The `vocalize` method
 // should return the pet's vocalization attribute.
 
-/* your answer here */
+const speak = function () {
```
benjimelito, line 83: I think you reversed these a little bit. What you wanted to do was add a `vocalization` property above, something like `this.vocalization = vocalization`. Below, you would add a method that returns this `vocalization` like so: `vocalize: function() {return this.vocalization}`

If I wanted to use this constructor, it would look something like this: `let petExample = new Pet('Spot', 'Dog', 'Terrier', 'Woof')`. Then if I wanted to call `petExample.vocalize()`, it would return `Woof`

benjimelito:
Solid attempt, just a few things to watch out for in here. 2/5

PR #228, opened 2017-03-27
Fallon diagnostic response

jonifallon:
I used my notes but I think I did better this time.  I wasn't exactly sure about the last question.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -17,7 +17,12 @@
 // Attributes: name, commissionRate
 // Methods: sell, call
 
-// Answer Starts Here
+Entity:  calendar
+Attributes: year, month, day
+Methods:  add, remove, update
+Entity:  events
+Attributes:  type, duration, places
+Methods: accept, create, share
 //
 // Answer Ends Here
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,20 +44,34 @@ let owner = null;
...
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name:  "Penelope",
+  species: "cat",
+  breed:  "Maine Coon",
+  vocalization: "meow",
+  Vocalize: function {
+    this.vocalization ++1
```
benjimelito, line 53: We don't want to increment `vocalization` here, we simply want to return it. Something like this: `function() {return this.vocalization}` so that when we call `pet.Vocalize()`, "meow" is returned.

lib/diagnostic.js
```diff
@@ -39,20 +44,34 @@ let owner = null;
...
+  this.breed = breed
+  this.vocalization = vocalization
+}
+
 
 // 5. Attach a `vocalize` method to the Pet's prototype.  The `vocalize` method
 // should return the pet's vocalization attribute.
 
-/* your answer here */
+this.vocalize = vocalize
```
benjimelito, line 74: All you want to do here is to add a method called `vocalize` to your `Pet` constructor. Something like this: `Pet.prototype.vocalize = function() {return this.vocalization}`

benjimelito:
Nice work, 3/5

PR #229, opened 2017-03-27
Responses for JS Object diagnostic

danielleletarte:
Comfort: 4
Clarity: 4

I used the previous exercise we did for "runs" where we created constructors and prototypes.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,13 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+//Entity: Calendar User
+//Attributes: name, calendar types (home, work)
+//Methods: add new calendar type, update profile
+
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,20 +49,33 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: 'Blitz',
+  species: 'cat',
+  breed: 'flame-point himmy',
+  vocalization: 'loud meows'};
```
benjimelito, line 57: We wanted to add a method directly here called `vocalize`, that would look something like this: `function() {return this.vocalization}`

benjimelito:
Nice! 4/5

PR #233, opened 2017-03-27
JS Object Diagnostic

cpm5041:
Comfort: 4
Clarity: 3

I felt relatively comfortable until the prototype question. I could
use more clarification and instruction on prototypes.

I used google and the hero exercise with this.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,14 @@
 // Methods: sell, call
 
 // Answer Starts Here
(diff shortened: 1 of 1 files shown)


benjimelito:
Everything looks good to me. Prototypes are a little bit tricky, but basically they are an object that contains methods and properties for all of its child objects to inherit from. Feel free to reach out if you'd like a more in-depth clarification. 5/5

PR #236, opened 2017-03-27
Add responses

lisamsmith100:
by Lisa Smith
linter was complaining quite a bit here

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -1,6 +1,11 @@
 'use strict';
 
-// Write your answers inside this file, at the places where it's indicated by
+//Comfort: 3
+//Clarity: 3
+
+//I used Google and my class notes to help with this diagnostic.
+
+//Write your answers inside this file, at the places where it's indicated by
 // the comments.
 
 // 1. Suppose that we wanted to create a browser-based calendar program for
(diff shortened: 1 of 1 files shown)


lisamsmith100:
I see what I did re: diff between literal and constructor type.  Just wanted to put that out there.

lib/diagnostic.js
```diff
@@ -39,24 +57,35 @@ let owner = null;
...
 
 // 5. Attach a `vocalize` method to the Pet's prototype.  The `vocalize` method
 // should return the pet's vocalization attribute.
 
-/* your answer here */
+Pet.prototype.vocalize = function () {
+  let noise = this.vocalization
+  }
+  return noise
+}
```
benjimelito, line 84: Nice! Although you don't have to define a new variable called `noise`, you can simply have your function return `this.vocalization`

lib/diagnostic.js
```diff
@@ -39,24 +57,35 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = new Pet()
+pet.name = 'Barbra Ann'
+pet.species = 'kitty'
+pet.breed = 'Tortoiseshell'
+pet.vocalization = 'muh muh muh muh muh muh meow'
```
benjimelito, line 64: Looks good, although you forgot to add a `vocalize` method to your object

benjimelito:
Good work, 4/5

PR #239, opened 2017-03-27
Js-objects-diagnostic

brianedmunds:
Responses to the diagnostic.

Comfort: 4
Clarity: 4

I used my notes for this exercise.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,16 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+
+Description: Browser-based calendar program for keeping track of things.
+Entity: Calendar
+Attributes:  time, day, week, month, year
+Methods: search, save, import, export
+
+Entity: Event
+Attributes: location, invitee, note, attachment
+Methods: alert, repeat, reminder
+
 // Answer Ends Here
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,24 +52,40 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+
+const useVocalization = function () {
```
benjimelito, line 56: More common to define this function inside the object, but this works too

lib/diagnostic.js
```diff
@@ -39,24 +52,40 @@ let owner = null;
...
+  this.vocalization = vocalization
+}
 
 // 5. Attach a `vocalize` method to the Pet's prototype.  The `vocalize` method
 // should return the pet's vocalization attribute.
 
-/* your answer here */
+const useVocalization = function () {
+  return vocalization
+}
```
benjimelito, line 84: You want to attach this to the prototype, not the object itself. Something like this: `Pet.prototype.useVocalization = function(){return this.vocalization}`. Let me know if you need some more clarification on how to use prototypes.

benjimelito:
Good work, 4/5

PR #240, opened 2017-03-27
Response to diagnostic

donpowers:


lib/diagnostic.js
```diff
@@ -39,21 +47,31 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: 'lucky',
+  species:'cat',
+  breed:'bobcat',
+  vocalization: 'meow'
+};
```
benjimelito, line 55: Nice, but you forgot to add a `vocalize` method to this object. Something like this: `vocalize: function() {return this.vocalization}`

benjimelito:
Nice work, 4/5

PR #241, opened 2017-03-27
Add Response

khan-ngo:
3/5

I'm getting it ... slowly

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,15 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+
+Entity: Owner
+Attributes: name, email
+Methods: add notes, update notes, remove notes
+
+Entity Two: calendar
+attributes: Year, month, daysOfWeek
+Methods: holds notes
+
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,20 +51,33 @@ let owner = null;
...
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: 'cuttie',
+  species: 'collie',
+  breed: 'arabian white'
+  vocalization: 'meowwww',
+  vocalize: function() {
+    console.log('purrrrrrr')
```
benjimelito, line 60: Ideally you want this method to reference the `vocalize` property that you defined above. Something like this: `return this.vocalization`

lib/diagnostic.js
```diff
@@ -39,20 +51,33 @@ let owner = null;
...
+  this.species: species,
+  this.breed: breed,
+  this.vocalization: vocalization
 };
 
 // 5. Attach a `vocalize` method to the Pet's prototype.  The `vocalize` method
 // should return the pet's vocalization attribute.
 
 /* your answer here */
+const vocalize.prototype = Vocalize(meoww)
```
benjimelito, line 80: You want to be adding this the `Pet prototype`, so something like this: `Pet.prototype.vocalize = function() {return this.vocalization}`

benjimelito:
Good job, but please try to get these in before 9:30am!

PR #243, opened 2017-03-27
Response for js-objects-diagnostics

rlarsonil71:
Comfort: 4
Clarity: 4

Used class notes and trainings for constructors, objects, and prototypes

Added the following to pet entity after commit to make grunt test work:
  vocalize: function () {
    return this.vocalization
  } 

let pet = {
  name: 'Marmaduke',
  species: 'Dog',
  breed: 'Siberian Husky',
  vocalization: 'Gentle',
  vocalize: function () {
    return this.vocalization
  }
};

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,16 @@
 // Methods: sell, call
 
 // Answer Starts Here
+// Description: A browser-based calendar program to keep track of things.
+// Entity 1:  Event
+// Attributes:  Event title, Event calendar (i.e. work, home), start day,
+//  start time, end day, end time, location, reoccurrence, reminderNotification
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,20 +52,30 @@ let owner = null;
 // Methods: vocalize
 
 // Replace null with your answer
-let pet = null;
+let pet = {
+  name: 'Marmaduke',
+  species: 'Dog',
+  breed: 'Siberian Husky',
+  vocalization: 'Gentle',
+};
```
benjimelito, line 60: I think you forgot to add a method here that returns your `vocalization` property. Something like this would work: `vocalize: function() {return this.vocalize}`

benjimelito:
Nice work, please try to get these in by 9:30am though!

PR #244, opened 2017-03-27
Completed Diagnostic

Ryaong792:
fist of 5 would be a 2 maybe a 3.  I understand the concept but was
not sure how to put it code.  Once i started it went ok but takes
longer for me to code it in.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,17 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+const User = {
+  Name = 'Ryan Ongchoa',
+  email = 'ryan.ongchoa@libertymutual.com',
+  meals = [
+    {
+    breakfeast: "shake w/ banna"
+    lunch: "salad"
+    dinner: "steak"
+  }
+],
+}
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -18,7 +18,17 @@
...
 // Answer Starts Here
-//
+const User = {
+  Name = 'Ryan Ongchoa',
+  email = 'ryan.ongchoa@libertymutual.com',
+  meals = [
+    {
+    breakfeast: "shake w/ banna"
+    lunch: "salad"
+    dinner: "steak"
```
benjimelito, line 28: This is an object, so make sure that all your properties are separated by a comma.

lib/diagnostic.js
```diff
@@ -38,21 +48,31 @@ let owner = null;
...
-// Replace null with your answer
-let pet = null;
-
+// Replace null with your answe
+let Pet = {
+  name = "blue Jay",
+  species = "bird",
+  breed = "Blue male jay bird....",
+  vocalization = "Raor!!!!",
+  }
```
benjimelito, line 57: Looks good, but I think you forgot to add a method that returns your `vocalization`. Something like this would work: `vocalize: function () {return this.vocalization}`

benjimelito:
Nice job, please try to get these in by 9:30am though!

PR #248, opened 2017-03-31
017/solution

BenGitsCode:
Refresh solution for cohort 017


MicFin (changes requested):
Solution is 1 commit behind and 2 commits ahead of master.  Must be 1 commit behind and 1 commit ahead.

BenGitsCode:
It's two commits ahead because it has the add solution and I ran the make standard on the solution, as it wasn't on master (obviously). Would you like me to squash those into an add solution commit with a note that the solution was also made standard?

jrhorn424:
Yes, please squash and complete the solution update. Thank you.

jrhorn424 (approved):

BenGitsCode:
I've squashed the two commits together @MicFin  @jrhorn424 by complete

PR #249, opened 2017-04-03
Completed js-objects-diagnostic

pjliddy:
Updated diagnostics.js with the answers to the 5 questions.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,14 @@
 // Methods: sell, call
 
 // Answer Starts Here
+// Description: Calendar App Entities
 //
+// Entity: Event
+// Attributes: date, time, location, attendees
+// Method: schedule, invite, delete
+//
+// Entity: Day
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,20 +50,33 @@ const owner = null
...
+    return (this.vocalization)
+  }
+}
 
 // 4. Create a constructor function for pet objects, as described above. The
 // constructor function should be responsible for setting the values of `name`,
 // `species`, `breed`, and `vocalization`.
 
 const Pet = function () {
-  /* Your answer here */
```
benjimelito, line 49: Here we were looking for something that would take 4 parameters, and then assign those to the object that is returned. Something like this: 
`const Pet = function (name, species, breed, vocalization) {
  this.name = name
  this.species = species
  this.breed = breed
  this.vocalization = vocalization
}`

benjimelito:
Looking good, 4/5

PR #252, opened 2017-04-03
Diagnostic completed

Maggicorp:
Submit diagnostic

difficulty 3
comfort 3

I didn't use grunt to test it because I haven't used it before but I will make sure to learn how to do that today so I can use it for future diagnostics.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,15 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+/* The browser-based calendar would want to have:
+Entity1: Event
+Attributes: time, date, description
+methods: Set Alarm/ reminder, change items to complete, delete items
+
+Entity2: User
+Entity: username, email address
+methods: Sent event info to users email address, abiiltity to invite or add other users
+*/
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,20 +51,39 @@ const owner = null
 // Methods: vocalize
 
 // Replace null with your answer
-const pet = null
+const pet = {
+  name: 'Buster',
+  species: 'cat',
+  breed: 'unknown',
+  vocalization: 'meows',
+  vocalize: function() { return console.log(this.vocalization)}
```
benjimelito, line 59: You can use either just `console.log(this.vocalization)` or `return(this.vocalization)`

lib/diagnostic.js
```diff
@@ -39,20 +51,39 @@ const owner = null
...
 }
+//let meow = new Pet('m', 's', 's', 't')
+//console.log(meow)
 
 // 5. Attach a `vocalize` method to the Pet's prototype.  The `vocalize` method
 // should return the pet's vocalization attribute.
 
 /* your answer here */
+Pet.prototype.vocalize = function () {
+  const vocalize = this.vocalization
```
benjimelito, line 82: You don't actually have to specify this variable inside here, the function will know what `vocalization` is

benjimelito:
Nice work, let someone know if you're having trouble with grunt. 5/5

PR #257, opened 2017-04-03
Response to diagnostic

mangoceylon:
comfort : 4.5
clarity : 4.5

Used previous lesson on constructors as a reference and google

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,6 +7,7 @@
 // keeping track of things. Identify at least two entites such a program might
 // need and describe attributes and methods of each.
 //
+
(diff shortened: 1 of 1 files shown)


benjimelito:
Good answers for the questions you got to, 3/5

mangoceylon:
My answers aren't in the right order, but I think I have all of them there, and it passes 14/14 the grunt tests

benjimelito:
Sorry, didn't see that your answer to number three actually satisfies number 4 as well. However, you didn't leave an answer for number 5 (it still passes the tests because the method is on the constructor, but it is not attached to the prototype). Updated to 4/5

PR #258, opened 2017-04-03
Diagnostic Answers

bcem2004:
Answers to questions 1-4 (ran out of time).

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,11 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+//Description: Our app manages daily course assignments
+Entity: calendar
+Attributes: day, week, month
+Methods: complete assignment, turn-in assignment
+
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -39,14 +43,20 @@ const owner = null
 // Methods: vocalize
 
 // Replace null with your answer
-const pet = null
+const pet = {name: "Rover", species: "dog", breed: "Golden Retriever", vocalize: "woof"};
+sound = function(){return pet.vocalize};
```
benjimelito, line 47: Make sure this is actually inside the `pet` object.

lib/diagnostic.js
```diff
@@ -39,14 +43,20 @@ const owner = null
...
 
 // 4. Create a constructor function for pet objects, as described above. The
 // constructor function should be responsible for setting the values of `name`,
 // `species`, `breed`, and `vocalization`.
 
 const Pet = function () {
```
benjimelito, line 53: You want this function to accept these as parameters, so something like `const Pet = function(name, species, breed, vocalize){...`

benjimelito:
Nice work on the questions you got to, 2/5

PR #261, opened 2017-04-03
Add response to JS Objects diagnostic

nriser:
I compared my work to the class work we did in hero.js, and made sure
I had everything I needed. It seems like
'const useVocalize = function () { return this._vocalize } is
redundant, as I already add 'Pet.prototype.useVocalize = function () {
return this._vocalize } to create and add useVocalize. Is this
understanding correct?

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,17 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+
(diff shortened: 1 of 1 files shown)


benjimelito:
Yes you're correct, adding the method directly to the object and adding it to the prototype are identical. Let me know if you have any questions about the difference between the two. 5/5

PR #262, opened 2017-04-03
Pull request for this diagnostic js-objects-diagnostic

noobiwankenoobi:
I believe that I'm on the right track with most of the questions
Possibly a couple of syntax fixes to make in the first few questions.
I did not finish the last question. Need to get faster/more comfortable with prototypes.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -18,7 +18,19 @@
 // Methods: sell, call
 
 // Answer Starts Here
-//
+Description: Our app is a calendar program
+
+Entity: Year
+Attributes: contains all the months in one Year
+Methods: change to new year,
+
+Entity: months
+Attributes: contains all the days in one months
+Methods: Change to a 30 or 31 day month, or 28 day month,
+
+Entity: day
+Attributes: contains all events for that day
+Methods: add new event,
 // Answer Ends Here
 
 // 2. Create an object literal that lines up with the following `owner` entity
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -29,8 +41,14 @@
...
 // Replace null with your answer
-const owner = null
+const owner = [
 
+{
+  name: 'Rick',
+  address:'Earth C-137',
+  numberOfPets: [1, 'dog', 'snuffles']
+}
+]
```
benjimelito, line 51: Don't need these brackets here, this would make it an array with one item

lib/diagnostic.js
```diff
@@ -39,20 +57,37 @@ const owner = null
 // Methods: vocalize
 
 // Replace null with your answer
-const pet = null
+const pet = [
```
benjimelito, line 60: Same as above ^^

lib/diagnostic.js
```diff
@@ -39,20 +57,37 @@ const owner = null
...
+{
+  name: 'Snuffles',
+  species: 'dog'
+  breed: 'Terrier'
+  vocalization: 'bark', 'bork'
+}
+]
```
benjimelito, line 67: Don't forget to add the `vocalize` method here!

lib/diagnostic.js
```diff
@@ -39,20 +57,37 @@ const owner = null
...
 
-const Pet = function () {
+const Pet = {
   /* Your answer here */
+function Pet (name, species, breed, vocalization) {
+  this.name = name;
+  this.species = species;
+  this.breed = breed;
+  this.vocalization = vocalization
+}
```
benjimelito, line 80: Nice

benjimelito:
Nice work, just a few syntax errors. Let me know if you have more in-depth questions about prototypes. 3/5

PR #267, opened 2017-04-03
All questions answered

d00medman:
Comfort :4
Clarity: 5

Had a little trouble with the final question re:prototyping. However, this would not have been a problem if I had more time. Computer was extremely slow to launch atom, causing me to lose ~10 min which could have been used on the diagnostic itself. Frustrating, but not debilitating.


benjimelito:
You still got all the right answers and got the pull request in on time, so nice work. 5/5

PR #269, opened 2017-04-03
 JS Objects Diagnostic

JLisaW:


lib/diagnostic.js
```diff
@@ -39,21 +52,31 @@ const owner = null
...
+  breed:
+}
 
 // 4. Create a constructor function for pet objects, as described above. The
 // constructor function should be responsible for setting the values of `name`,
 // `species`, `breed`, and `vocalization`.
 
-const Pet = function () {
-  /* Your answer here */
+const Pet = function() {
```
benjimelito, line 65: Make sure to pass in `name, species, breed, vocalization` as parameters

benjimelito:
Make sure to get these in before 9:30am! Looks like you had a good start, but make sure to let someone know if you need any more help with this. 0/5

